PR #5, opened 2022-11-10
[AB#37116] Support added for additional licensing urls for entitlement webhook to enable video playback

AxTrusov:
3 new env variables added for different license services (one for each DRM technology) and one for Fairplay Certificate.
Configured values are returned by the `/entitlement` webhook together with the entitlement JWT.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -15,6 +15,7 @@
     "codegen",
     "Dockerized",
     "Downloadability",
+    "fairplay",
     "geolocation",
     "Inflector",
     "loglevel",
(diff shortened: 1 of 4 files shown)


.vscode/settings.json
```diff
@@ -15,13 +15,15 @@
     "codegen",
     "Dockerized",
     "Downloadability",
+    "fairplay",
```
Zehelein, line 18: Not sure about the spelling plugin - but correctly the term would be "FairPlay" and further down "PlayReady" and "Widevine".

Zehelein (approved):
Please check if the spelling settings would work better with the correct casing. Otherwise, this looks good. Thanks!

.vscode/settings.json
AxTrusov: This option is added using the VSCode lightbulb in all lowercase and checks spelling in all cases, so no need to adjust it here.

PR #7, opened 2022-11-15
[AB#29105] fix: :licensing-from/to same date not allowed

gaveen5555:
Currently it's possible to set the same date time to both fields (From and To date fields). It has been fixed and now a error message is shown.


services/media/workflows/src/Stations/Movies/MovieLicensingDetails/MovieLicensingDetails.tsx
```diff
@@ -32,15 +32,21 @@ type FormData = MutationUpdateMoviesLicenseArgs['input']['patch'] & {
...
+      if (value) {
+        if (parent.licenseStart.getTime() >= value.getTime()) {
+          return createError({
+            path,
+            message: 'License end date cannot be before start date',
+          });
+        }
+      }
+      return true;
+    }),
```
axisuru: This can be simplified as, 

```suggestion
    .test(
      'checkEndDate',
      'License end date cannot be before start date',
      function (value) {
        const { parent } = this;
        if (value) {
          return parent.licenseStart.getTime() < value.getTime();
        }
        return true;
      },
    ),
```

axisuru (approved):

PR #13, opened 2022-12-05
[AB#37340] Video and stream properties added and renamed

AxTrusov:
Media:
- requests to encoding and image services are now done with codegen components, providing stronger typing.
- when a video is published (in the context of a movie, episode, etc..), newly added and renamed video and stream properties are now used.
- yup validation adjusted for videos and images to accommodate usage of codegen components and updated types.
- generic error message for snapshot processing error adjusted in workflows.
	
Catalog:
- data migration was done to reflect recent changes to the encoding service, including adding new properties, renaming existing ones, and adjusting stored values where necessary
- because numeric values with fractional parts are now stored and exposed via API, a dedicated GraphQL plugin was added to use Float instead of BigFloat (which would resolve as a string by default)
- `jest-extended` dev dependency added to introduce more test assertions (e.g. for arrays comparison)
	
Entitlement and Libs:
- Adjustments to accommodate renamed properties
	
List of renamed properties:
- Video Stream:
	- drm_key_id -> key_id
	- initial_file -> file
	- bandwidth_in_bps -> bitrate_in_kbps
- Video
	- duration/duration_in_seconds -> length_in_seconds

List of added properties (for Video Stream):
- type
- file_template
- codecs
- frame_rate
- height
- width
- display_aspect_ratio
- pixel_aspect_ratio
- sampling_rate
- language_name
	
**Special note on the removal of `ignoreIndexes: false,` from the catalog service:**

The original intention of this flag is to disable filtering and ordering via GraphQL API for properties/columns that do not have indexes set on them on the database level. Unfortunately, this [flag is broken](https://discord.com/channels/489127045289476126/498852330754801666/918156014002528326) as stated by Benjie [multiple](https://discord.com/channels/489127045289476126/498852330754801666/931852284902637608) [times](https://discord.com/channels/489127045289476126/498852330754801666/864500534978150420). Reason: [`Basically it's not smart enough to understand partial indexes, indexes with operations and expressions, etc; and it's written on top of the "omit" functionality so it incorrectly removes relations in both directions whereas it should only remove one direction.`](https://discord.com/channels/489127045289476126/498852330754801666/918165255199285260)

In this particular PR, it was producing inconsistent behavior, failing to detect indexes on a table with existing rows, and disabling filtering on the `type` column of video stream tables when it should have allowed it.

As an alternative, filtering and ordering are now explicitly disabled via smart tags, and an extra `README.md` section was added to indicate that.


libs/media-messages/schemas/payloads/publish/common.json
```diff
@@ -139,21 +139,63 @@
...
+        },
+        "file": {
+          "type": ["string", "null"],
+          "description": "File path to the initialization segment"
         },
         "iv": {
-          "description": "IV",
-          "type": "string"
+          "type": ["string", "null"],
+          "description": "IV"
```
Zehelein: The initialization vector that was used to encrypt this media file.

services/media/service/src/domains/common/utils/get-images-metadata.ts
```diff
@@ -22,43 +23,28 @@ export const getImagesMetadata = async (
...
 
-    if (result.data.errors?.length > 0) {
+    if (!data.images?.nodes) {
       throw new MosaicError({
         ...CommonErrors.PublishImagesMetadataRequestError,
-        details: {
-          errors: result.data.errors,
+        logInfo: {
+          reason:
+            'Request to the image service succeeded, but no images were returned and explicit error was not thrown.',
```
Zehelein: The request to the Image Service succeeded, but no images were returned and an explicit error was not thrown.

services/media/service/src/domains/common/utils/get-images-metadata.ts
```diff
@@ -92,15 +78,18 @@ export const getImagesMetadata = async (
     return { result: publishData, validation };
   } catch (error) {
     const mapper = mosaicErrorMappingFactory(
-      (error: Error & { response?: { data?: { errors?: unknown[] } } }) => {
+      (error: Error & { response?: { errors?: unknown[] } }) => {
```
Zehelein: The mapper can be moved outside of the `getImagesMetadata` function.

Zehelein (commented):
Looks good - only some smaller comments.

Zehelein (approved):

PR #26, opened 2022-12-22
[AB#37536] Add documentation about production builds

BenjaminSchwendner:
# Pull Request

## Prerequisites

- [ ] All tests pass

## Description

Adding documentation regarding how to create production artifacts to the `README.md` and removing the GitHub workflows we recently added, as we will continue running our builds using Azure DevOps pipelines instead.
The documentation should allow users to setup their own build pipeline independent of the tooling they would like to use.


README.md
```diff
@@ -125,6 +134,63 @@ environment.
...
+```sh
+npx piral-cli@latest publish-pilet --no-fresh --url "https://frontends.service.eu.axinom.net/v1/pilets/{tenantId}/{environmentId}" --api-key "{serviceAccountCredentials}" {packageFile}
+```
+
+Please note, that you need to provide the 'Base64 Encoded Credentials' of a
+service account that has the `Pilets: Publish` permissions as `api-key`. This
+value will be shown to you alongside with the other secrets when creating a
+service account or when generating new secrets for an existing account.
+
+Also make sure that you define all values the pilet expects to receive inside
```
EmperorRXF, line 186: maybe we can hint here that these values are generally found in the .env.template file and should be used as KEY/VALUE pairs below?

README.md
```diff
@@ -125,6 +134,63 @@ environment.
...
+the `app.meta.custom` by providing them as additional parameters on this command
+like:
+
+```sh
+--fields.{key} {value}
```
EmperorRXF, line 191: should we give an actual example here (maybe the full npx command repeated again)?

EmperorRXF (approved):

README.md
BenjaminSchwendner: added a note

README.md
BenjaminSchwendner: considered that, but it felt a bit awkward to have basically the same command there again, just with the `--fields` argument added...
People can also check the Piral documentation about the command if there is something unclear (https://docs.piral.io/tooling/publish-pilet) or wait for our hosting service which will also change the way how the workflows (can) get deployed

EmperorRXF (approved):

PR #29, opened 2023-01-24
[AB#37922] add portal version check on workflow dev build

ingaeglite:
# Pull Request

## Prerequisites

- [ ] All tests pass

## Description

Adds a script that compares latest version of package `@axinom/mosaic-portal` from npm registry with package.json files
And outputs a message to update package if local version is older
Script is run on workflow dev build


scripts/package-version-check.ts
```diff
@@ -0,0 +1,62 @@
...
+ * It outputs message to update package if local version is older.
+ */
+async function checkPackageVersion(): Promise<void> {
+  let latestVersion = '';
+  try {
+    latestVersion = execSync(`yarn info ${packageName} version`)
+      .toString()
+      .trim();
+  } catch (error) {
+    // eslint-disable-next-line no-console
```
BenjaminSchwendner: we can even allow that globally for the whole file, as this is a script and it's intended to write to the console

ingaeglite: Disabled `no-console` for entire file

BenjaminSchwendner (approved):

scripts/package-version-check.ts
```diff
@@ -0,0 +1,61 @@
...
+      error ? reject(error) : resolve(files),
+    );
+  });
+
+  for (const packageJsonPath of packageJsonPaths) {
+    const packageJson = JSON.parse(fs.readFileSync(packageJsonPath, 'utf-8'));
+    const dependencies = packageJson.dependencies || {};
+    const devDependencies = packageJson.devDependencies || {};
+
+    if (dependencies[packageName] || devDependencies[packageName]) {
```
EmperorRXF: tiny optimization - could remove this check as it's anyway covered in L49

ingaeglite: Indeed it is not needed in the final version. Removed it.
Thanks for noticint!

EmperorRXF (approved):

PR #30, opened 2023-01-29
[AB#37925] Version media-workflows based on build number

EmperorRXF:
# Pull Request

## Prerequisites

- [x] All tests pass

## Description

With this PR, the media-workflow build pipeline will update the package version based on the build number. When the pilet is released, this will help to track the currently running workflow version on an environment

### DIFF
--- a/.adops/workflow-build.yml
+++ b/.adops/workflow-build.yml
@@ -7,6 +7,8 @@ variables:
   YARN_CACHE_FOLDER: $(Pipeline.Workspace)/.yarn
   # Agent VM image name
   vmImageName: 'ubuntu-latest'
+  # Workflow package path
+  workflowPkgPath: './services/media/workflows'
 
 stages:
   - stage: Build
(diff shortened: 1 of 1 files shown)


.adops/workflow-build.yml
```diff
@@ -30,6 +32,18 @@ stages:
...
+          - script: |
+              ORIGINAL_PKG_VER=`npx -q json -f "$(workflowPkgPath)/package.json" version`
+              NEW_PKG_VER="${ORIGINAL_PKG_VER}-$(tag)"
+
+              echo "Current package.json version is: [${ORIGINAL_PKG_VER}]"
+              echo "Updating package.json version to: [${NEW_PKG_VER}]"
+
+              npx -q json -I -f "$(workflowPkgPath)/package.json" -e "this.version = '${NEW_PKG_VER}'"
+
+              echo "Updated package.json version is: [`npx -q json -f "$(workflowPkgPath)/package.json" version`]"
```
yohanAnushka: @EmperorRXF I think we gotta escape the quotations after `json -f` part. Otherwise it encompasses the rest of the yaml including `displayName` field.

EmperorRXF: seems the issue is with GH's web-based syntax highlighter not being too great at parsing nested double quotes. it fails to parse even with the escapes :)

but I will still leave in the escaped quotes, as the variable expansion on bash works nevertheless (and semantically since it's a nested string I don't mind keeping the inner ones escaped)

yohanAnushka (approved):
LGTM

PR #31, opened 2023-02-01
[AB#29105] Licensing start and end date validation now uses minutes instea…

gaveen5555:
…d of miliseconds

# Pull Request

## Prerequisites

- [ ] All tests pass


BenjaminSchwendner:
why this change? `getMinutes` will only return the minutes that passed in the given hour. Having two dates from different hours will result in unexpected/random results.
If I don't miss anything here, `getTime` was the better option to compare the two dates.

If the problem was, that we also consider dates the same at the moment the minutes are the same, the solution would rather be, to "remove" the milliseconds and seconds from the `getTime` value  before we compare the values.

BenjaminSchwendner:
https://github.com/Axinom/mosaic-media-template/pull/41

PR #34, opened 2023-02-13
[AB#37460] Added support for DRM protected videos and enabled DRM protection of the Virtual Channel

IrinaBazyleva:
# Pull Request

## Prerequisites

- [X] All tests pass

## Description

VOD-to-Live Service was extended  with next functionalities:
- provide to the Virtual Channel API CPIX decryption files, that allows Unified Streaming to create Live Streams from DRM encrypted videos.
- provide CPIX protection files, that enable Virtual Channel Live Stream protection.


services/vod-to-live/service/src/domains/cpix/generator/generate-cpix-settings.ts
```diff
@@ -0,0 +1,109 @@
...
+export const generateCpixSettings = async (
+  videos: DetailedVideo[],
+  storage: AzureStorage,
+  keyServiceApi: KeyServiceApi,
+  settingAccessStartDateTime: Date,
+  settingAccessDurationInSeconds: number,
+  channelId: string,
+  playlistId?: string,
+): Promise<CpixSettings> => {
+  const h = 1; //number of hours added to expiration date end
```
Zehelein: I would either _not_ add any additional time _or_ rename the settingAccessDurationInSeconds to something like "minimum...". But in the end, I would prefer to set the exact time from the caller's side. It should likely be the duration of the playlist plus the configured catch-up time.

services/vod-to-live/service/src/domains/cpix/generator/generate-cpix-settings.ts
```diff
@@ -0,0 +1,109 @@
...
+  const durationOfSasInMilliseconds =
+    settingAccessDurationInSeconds * 1000 + h * 60 * 60 * 1000;
+  const decryptionSasEndDate = new Date(
```
Zehelein: Why does the decryption have a shorter date than the encryption? Wouldn't it require to be the same in order to decrypt with the VOD DRM key and then protecting with the live output key?

services/vod-to-live/service/src/domains/live-stream/prepare-transition-live-stream.ts
```diff
@@ -85,23 +85,15 @@ const createTransition = async (
...
-const isFutureDate = (date: string): boolean => {
-  const today = new Date();
-  const futureDate = new Date(date);
-  return futureDate > today;
-};
-/**
- * Creates a ISO date time for the transition start
+ * Creates a ISO date time for the transition start.
+ * If playlist start date is in future - the start date is used.
+ * Otherwise, transition date time is calculated by formula Now + 5 minutes.
```
Zehelein: Where do the 5 minutes come from? Is it the time it takes for the Unified Streaming solution to process a playlist? And if 5 (or whatever amount of) minutes are required - then the "isFutureDate" should likely also take this into consideration or?

services/vod-to-live/service/src/domains/validation/utils.ts
```diff
@@ -11,10 +11,12 @@ export const validateVideo = (video: DetailedVideo): ValidationResult => {
     warnings: [],
   };
   if (video.video_encoding.is_protected) {
-    validationResult.errors.push({
-      message: `Video ${video.id} is DRM protected.`,
-      code: 'VIDEO_IS_PROTECTED',
-    });
+    if (video.video_encoding.video_streams.find((s) => !s.key_id)) {
+      validationResult.errors.push({
+        message: `Video ${video.id} is missing DRM Keys.`,
```
Zehelein: The encoding service only manages the DRM key IDs - so we should also adjust the message and code to "key id".

services/vod-to-live/service/src/setup/create-protection-cpix.ts
```diff
@@ -0,0 +1,54 @@
...
+  createDashCpixRequest,
+  createHlsCpixRequest,
+  KeyServiceApi,
+  protectionDashCpixFileName,
+  protectionHlsCpixFileName,
+  storeSpekeResponse,
+} from '../domains';
+
+/**
+ * Executed once on service startup: creates in the root of the Azure storage
```
Zehelein: * Executed once on service startup: creates in the root of the Azure storage container
 * the CPIX files that are required for DASH and HLS live stream protection.
 * All live streams created by the service will use files for protection
 * with Azure SAS token limited in time by the playlist/channel duration.

Zehelein: And I would add a note that this can be further extended to create new CPIX files for every channel or for every published playlist.

Zehelein (changes requested):

services/vod-to-live/service/src/common/constants.ts
```diff
@@ -0,0 +1,4 @@
+export const DAY_IN_SECONDS = 86400;
+export const HOUR_IN_SECONDS = 3600;
+export const SECOND_IN_MILLISECONDS = 1000;
+export const MINUTE_IN_MILLISECONDS = 60 * SECOND_IN_MILLISECONDS;
```
Zehelein, line 4: Why do you calculate this but not the others? 😀

services/vod-to-live/service/src/domains/live-stream/prepare-channel-live-stream.ts
```diff
@@ -38,20 +52,37 @@ export const prepareChannelLiveStream = async (
...
+        channelKey.Id,
+        channelId,
+        keyServiceApi,
+        storage,
+      );
       const channelCreationResult = await virtualChannelApi.putChannel(
         channelId,
         smil,
       );
+      //TODO: send RabbitMQ event message to channel's key_id (or send channel metadata + key_id?)
```
Zehelein: What is the TODO here? Send which message to a channel key id (would that be a special routing key)?

services/vod-to-live/service/src/domains/live-stream/prepare-channel-live-stream.ts
```diff
@@ -63,3 +94,45 @@ export const prepareChannelLiveStream = async (
     throw getMappedError(error);
   }
 };
+
+/**
+ * Executed once per each new channel: creates in the Azure storage in directory {channelId}
+ * CPIX files required for DASh and HLS live stream protection.
```
Zehelein: DASh --> DASH

Zehelein (commented):
Looks good - added only small comments 👍

PR #42, opened 2023-02-27
[AB#38605] Empty key ids are now excluded from entitlement message

AxTrusov:
[[AB#38605](https://dev.azure.com/axinom/5e61ad7f-f562-45f3-8078-9ade7df639c1/_workitems/edit/38605)]

# Pull Request

## Prerequisites

- [x] All tests pass

## Description

Subtitle and Closed Caption streams do not have a keyId values, but were previously included during the generation of the entitlement message. Empty keyId values are now excluded.

### DIFF
--- a/services/entitlement/service/src/graphql/plugins/entitlement-endpoint/catalog-service-communication.ts
+++ b/services/entitlement/service/src/graphql/plugins/entitlement-endpoint/catalog-service-communication.ts
@@ -86,10 +86,11 @@ const getCatalogResponseVideo = (
     });
   }
 
+  const [video] = videos;
   if (
-    !videos[0].isProtected ||
-    !videos[0].videoStreams ||
-    (videos[0].videoStreams?.nodes ?? []) // A protected video will always have at least one stream with non-empty Key ID
+    !video.isProtected ||
+    !video.videoStreams?.nodes?.length ||
+    video.videoStreams.nodes // A protected video will always have at least one stream with non-empty Key ID
       .map((videoStream) => videoStream.keyId)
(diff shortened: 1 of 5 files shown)


services/entitlement/service/src/graphql/plugins/entitlement-endpoint/catalog-service-communication.ts
```diff
@@ -89,8 +89,7 @@ const getCatalogResponseVideo = (
   if (
     !videos[0].isProtected ||
     !videos[0].videoStreams ||
-    videos[0].videoStreams.nodes.length === 0 || // A protected video will always have a video_streams record
-    videos[0].videoStreams.nodes
+    (videos[0].videoStreams?.nodes ?? []) // A protected video will always have at least one stream with non-empty Key ID
```
Zehelein: Just as a discussion point: not sure if I would remove the length check. If anything it acts as a TypeScript save-guard and the line 92 is more confusing to me on why you would map over an empty array. So the old variant is more explicit - even if in both cases it really cannot happen it is a nice safety net.

Zehelein (approved):
Added a minor comment about which code adds more clarity. Otherwise it looks good.

services/entitlement/service/src/graphql/plugins/entitlement-endpoint/catalog-service-communication.ts
AxTrusov: I was trying to combine length and null/undefined checks. Refactored this part of the code a bit so it would still include the length check, not include the `?? []`, but would still make existing tests pass.

Zehelein (approved):

PR #55, opened 2023-03-24
[AB#37767] Upgrade lib-jsonwebtoken to 9.x to fix some security vulnerabilities

axinom-nrathnayaka:
# Pull Request

## Prerequisites

- [x] All tests pass

## Description

Upgrade lib-jsonwebtoken to 9.0.0


EmperorRXF (changes requested):
think we would need to update the new RC versions of the mosaic libs as well?

EmperorRXF:
we opened a separate PR to handle this change #56 , and this is now redundant

PR #66, opened 2023-05-08
[AB#38392] [AB#39185] [AB#39587] [AB#39035] [AB#37460] [AB#36848][AB#38416][AB#40494] VOD-to-Live service

IrinaBazyleva:
# Pull Request

## Prerequisites

- [x] All tests pass

## Description

Contains new customizable service VOD-to-Live. The service implements custom logic and creates channels live streams using the
[Unified Virtual Channel](https://beta.docs.unified-streaming.com/documentation/virtual-channel/index.html).

## Testing notes:

To perform testing, the following services are required:

**Core Services:** All Mosaic core services, including Id, FE, Hosting, and
others.

**Managed Services:** Channel Service, Video Service, and Image Service.

**Customizable Services:** VOD-to-Live Service, Media Service, and Catalog
Service.

**3rd party:** USP Virtual Channel API

Before testing, please ensure that the `/pre-publishing` webhook to the
VOD-to-Live Service is correctly configured in the Channel Service configuration
in the Admin Portal.

### Content Preparation

In preparation for testing the VOD-to-Live Service, ensure that videos are available to assign as Placeholder Videos for the Channels, and that Media Assets (movies and episodes) are prepared for adding to the playlist programs. All videos used in testing should be encoded using the Content Processing profile, with the following settings: output format - CMAF, and DRM protection disabled (DRM protected videos can be used, only if VOD-to-Live service has DRM protection enabled).

```
Note all videos associated with the same Channel (including the videos associated with Media Assets added to playlist)
should have at least one common stream format (matching resolution, frame rate, and bitrate).
```

Please add cue points to the videos associated with the Media Assets before adding them to the Playlist. Otherwise, the cue points will not be registered in the Channel Service.

### Create a Live Stream for the Channel:

In the Channel Service UI:

1.  Begin by creating a new channel and filling in the required metadata. Then, add the channel logo and placeholder video.
2.  Navigate to the channel details page and click the "Publishing" button. This will redirect you to a new page for validation.
3.  If the VOD-to-Live service pre-publishing webhook is properly configured,  the validation page should display validation results from both the Channel Service and VOD-to-Live service.
4.  If any validation errors appear, the channel publication will be blocked. Please address all errors before proceeding.
5.  Once the validation passes successfully, click the "Publish" button and confirm the channel publication.

When you perform the "Publish" action, a RabbitMQ message containing the channel metadata is sent. One copy of this message is received by the Catalog API, and the published channel metadata becomes available in the channels endpoints of the Catalog Service. However, the URLs where the channel can be played will not yet be available at this stage.

A second copy of the Channel message is received by the VOD-to-Live service. This service transforms the JSON object from the message into the SMIL format and requests the creation of the Channel live stream from the USP Virtual Channel API. After requesting the creation of the channel live stream, the VOD-to-Live service will periodically ping the USP API while waiting for the Channel Live stream to become live(service waits for the channel to go live for a configured amount of time, by default 10 minutes). When a live stream of the Channel becomes available for playback VOD-to-Live will send a RabbitMQ message to the Catalog API, containing URLs for the Channel live stream.

6.  The Catalog GraphQL should contain endpoints for Channels. Channel information should contain dashStreamUrl and hlsStreamUrl. These URLs can be used in online players such as Shaka, dash.js, and hls.js to enable playback.
7.  If only Channel data is published from the Channel Service, the live stream created by USP will display the Channel Placeholder video on repeat.

### Add a Playlist schedule to the Channel Live stream:
In the Channel Service UI:
1.  Create a new playlist for the Channel published in previous step. Set the Playlist start date to the reasonable date and time in the future. 
2. Create a schedule for the playlist by navigating to the "Program". 
3. Add Media Asset to playlist program and fill-in video cue points with AD_PODs and trailers.
4. When Program is ready, from the Playlist details station press 'Publishing' button.
5. Similarly to the Channel, a validation station for the Playlist will show validations from Channel Service and VOD-to-Live service. If any validation errors appear, the Playlist publication will be blocked. Please address all errors before proceeding.
6. Once the validation passes successfully, click the "Publish" button and confirm the Playlist publication.

Similar to the Channel Publish action, a RabbitMQ message containing the Playlist metadata is sent when a Playlist is published. This message is received by the VOD-to-Live Service, which transforms the JSON object from the message into the SMIL format. The service then requests the creation of a transition for the Channel USP Virtual Channel API. The Virtual Channel Transition is a set of rules that dictate when and what content should be played in the live stream at a particular time.

When a Playlist is published, the transition for the Virtual Channel will start at the published Playlist's start time. All programs assigned to the Playlist will be the content played from that time onwards on the Virtual Channel.

7. When the Channel's live stream (URLs received in the 'Create a Live Stream for the Channel' section) is being played at the scheduled time for the published Playlist, the live stream content should change from the Channel Placeholder playback on a loop to the Media Assets assigned to the Playlist.
8. As there is no integration with SSAI (server-side ad insertion), instead of ads at the cue points filled as 'AD_POD's, the Channel Placeholder video on a loop will be played.


.vscode/settings.json
```diff
@@ -12,6 +12,8 @@
   "cSpell.words": [
     "Axinom",
     "Axinom's",
+    "CMAF",
+    "cpix",
```
Zehelein: It does not matter much for the plugin but would likely be best to keep the casing consistent. So acronyms like CMAF should be upper case IMO

libs/media-messages/schemas/payloads/vod-to-live/commands/ensure-channel-live-command.json
```diff
@@ -0,0 +1,21 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "title": "ensure_channel_live_command",
```
Zehelein: The command name is a bit confusing (to me). Could it be renamed to something like "start_channel_command" or "start_stream_channel_command" or something similar?

services/vod-to-live/service/.env.template
```diff
@@ -0,0 +1,47 @@
+NODE_ENV=development
+SERVICE_ID=vod-to-live-service
+LOG_LEVEL=DEBUG
+
+PORT=11900
+
+# These variables are optional and are only needed for local development
```
Zehelein: Only the next line? If yes then we should use singular.

services/vod-to-live/service/.env.template
```diff
@@ -0,0 +1,47 @@
...
+
+# Key seed id, used for encryption keys generation
+DRM_KEY_SEED_ID=
+
+# If `PROLONG_PLAYLIST_TO_24_HOURS` is set to true, this duration is added on top of the 24h mark for smoother transition between playlists.
+# If `PROLONG_PLAYLIST_TO_24_HOURS` is set to false, the catch up duration is added to the playlist transition time, when playlist start date is in the PAST.
+# Defaults to 1 hour (60 minutes)
+CATCH_UP_DURATION_IN_MINUTES=60
+
+# Feature flag. When turned on every playlist with duration under 24 hours
```
Zehelein, line 33: maybe move this section above the other to know what this is about when you first read it in line 35

services/vod-to-live/service/README.md
```diff
@@ -0,0 +1,250 @@
...
+## Overview
+
+The VOD-to-Live service is a customizable service that is part of the Media
+solution. The service implements custom logic and creates live streams using the
+[Unified Virtual Channel](https://beta.docs.unified-streaming.com/documentation/virtual-channel/index.html).
+
+The managed Channel service provides the possibility to publish Channels and
+Playlists to the customizable VOD-to-Live service. The VOD-to-Live service uses
+the Channel service data to create a Synchronized Multimedia Integration
+Language (SMIL) file with the videos from the playlist and “empty” videos as
```
Zehelein: "empty" videos might be hard to understand. Basically an editor selects a video that should be used as a placeholder/filler video when no planned playlist is streamed.

services/vod-to-live/service/README.md
```diff
@@ -0,0 +1,250 @@
...
+- All videos associated with the Playlist should have at least one mutual stream
+  format (have matching resolution, frame rate and bitrate)
+
+#### Video validation rules
+
+```
+Every video that is assigned to a channel or playlist undergoes validation based on the following set of rules.
+```
+
+- The Video codec should be `H264`.
```
Zehelein: Maybe a bit more consistent to write it like this:

- The Video should be encoded as `H264`.
...
- The Video should be packaged in the `CMAF` format.

services/vod-to-live/service/README.md
```diff
@@ -0,0 +1,250 @@
...
+
+```
+
+#### DRM protection
+
+The Channel's live stream can be DRM protected. To enabled the DRM protection
+adjust the service configuration by setting the DRM variables.
+
+```
+Service supports only the Axinom DRM Key Service.
```
Zehelein: > The service uses the Axinom DRM Key Service to request the DRM encryption key.

services/vod-to-live/service/README.md
```diff
@@ -0,0 +1,250 @@
...
+DRM settings for the VOD-to-Live service should align with the DRM settings used for the encoding of the VODs.
```
Zehelein: > for the encryption of the used VOD assets.

services/vod-to-live/service/README.md
```diff
@@ -0,0 +1,250 @@
...
+by time.
+
+For each Channel will be created new DRM key through the Axinom Key Service API.
+The key identifier of the channel will be send in RabbitMQ message
+`LiveStreamProtectionKeyCreatedEvent`, identifier is stored with the Channel's
+metadata in Azure Storage.
+
+#### Playlist prolongation
+
+When the Playlist prolongation feature is enabled, X Service automatically
```
Zehelein: what is "X Service"?

services/vod-to-live/service/scripts/setup-webhook.ts
```diff
@@ -0,0 +1,123 @@
...
+const mapAndLogError = mosaicErrorMappingFactory(
+  (error: Error & { code?: string; config?: { url?: string } }) => {
+    if (error?.code === 'ECONNREFUSED') {
+      console.log(
+        'Unable to make a request to the following URL because the service is not accessible:',
+        error?.config?.url,
+      );
+      return undefined;
+    }
+    console.log(
```
Zehelein, line 26: Please don't use `console`. Maybe check if it is anyways logged via the base error mapping factory or create a new logger. Hm - as it is in the scripts folder this might not be that important but still might be a good thing to always do.

services/vod-to-live/service/src/domains/azure/azure-blob-errors.ts
```diff
@@ -0,0 +1,22 @@
+export const AzureBlobErrors = {
+  AuthenticationFailed: {
+    message: 'Unable to authenticate access to the Azure Storage.',
+    code: 'AZURE_BLOB_AUTHENTICATION_FAILED',
+  },
+  ContainerNotFound: {
+    message: 'The configured container does not exist.',
```
Zehelein: + "Azure Storage" (just in case to make sure if somebody just reads the title to know the corresponding area). Also for line 19

services/vod-to-live/service/src/domains/cpix/generator/generate-cpix-settings.ts
```diff
@@ -0,0 +1,129 @@
...
+ * Generates a CpixSetting, containing paths to CPIX files required for videos decryption and Live Stream protection.
+ * @param channelId - unique identifier of the Channel.
+ * @param playlistId - playlist unique identifier, if applicable.
+ * @param decryptionParams - parameters for decryption of DRM protected videos assigned to stream.
+ * @param encryptionParams - parameters for protection of resulting Live Stream.
+ * @param storage  - reference to Azure Storage class instance.
+ * @param keyServiceApi - reference to Key Service API class instance.
+ * @returns
+ */
+export const generateCpixSettings = async (
```
Zehelein: With the "single" responsibility: could you split this function into create encryption and decryption cpix functions (and the encrypt function would be called twice - once for dash, once for hls)?

services/vod-to-live/service/src/domains/smil/generator/channel-smil-generator.ts
```diff
@@ -0,0 +1,82 @@
...
+    if (originalEvent.placeholder_video) {
+      const placeholderVideo = videoToSmilParallelReferences(
+        originalEvent.placeholder_video,
+        extractSharedVideoStreamFormats([originalEvent.placeholder_video]),
+      );
+      parallels.push(createParallel(placeholderVideo));
+    } else {
+      // Virtual Channel cannot be created without the SMIL with a valid parallel/video
+      throw new MosaicError({
+        message: `Channel ${originalEvent.id} is missing placeholder video. Virtual Channel cannot be created.`,
```
Zehelein, line 39: message: `The channel with ID ${originalEvent.id} is missing the placeholder video. Therefore the virtual channel cannot be created.`,

services/vod-to-live/service/src/domains/smil/generator/playlist-smil-generator.ts
```diff
@@ -0,0 +1,395 @@
...
+        ...this.tryProlongatePlaylist(
+          originalEvent.start_date_time,
+          originalEvent.end_date_time,
+        ),
+      );
+    }
+
+    // if first and last parallels are `out of network`
+    // come back 'in network' parallel should be added for proper looping in USP
+    const firstParallel = parallels[0];
```
Zehelein: Index position 1 is not guaranteed to exist if line 117 did fall back to an empty array and line 121 was false.

services/vod-to-live/service/src/domains/smil/generator/playlist-smil-generator.ts
```diff
@@ -0,0 +1,395 @@
...
+    const postCuePoint = program.program_cue_points?.filter(
+      (cp) => cp.type === 'POST',
+    )[0];
+
+    // adding PRE cue point
+    if (preCuePoint) {
+      parallels.push(...this.processCuePoint(preCuePoint));
+    }
+
+    //cutting program video for MID cue points
```
Zehelein, line 223: Please introduce to eslint `"spaced-comment": ["error", "always"],` to have consistent comment styling (agreed with Benny). Should be auto fixable according to their docu https://eslint.org/docs/latest/rules/spaced-comment (likely also in other places)

services/vod-to-live/service/src/domains/smil/generator/playlist-smil-generator.ts
```diff
@@ -0,0 +1,395 @@
...
+                  ? undefined
+                  : result.lastSpliceTimeInSeconds,
+                midCuePoint.time_in_seconds,
+              ),
+            );
+            result.lastSpliceTimeInSeconds = midCuePoint.time_in_seconds;
+            result.parallels.push(...this.processCuePoint(midCuePoint));
+          } else {
+            this.logger.warn({
+              message: `MID cue point '${midCuePoint.id}' is missing insertion time. Cue point is skipped! `,
```
Zehelein: > The MID cue point with ID '${midCuePoint.id}' is missing the insertion time. Therefore this cue point is skipped!

Please adjust similar places as well.

services/vod-to-live/service/src/domains/smil/generator/utils.ts
```diff
@@ -0,0 +1,202 @@
...
+    );
+  }
+  return adPlaceholderParallels;
+};
+
+export const determineIntegerDivisionAndRemainder = (
+  number: number,
+  divisor: number,
+): { quotient: number; remainder: number } => {
+  return {
```
Zehelein, line 107: Likely best to check that divisor is larger than zero - if not return e.g. quotient: 0 and remainder 0

services/vod-to-live/service/src/domains/validation/playlist-published-validation-handler.ts
```diff
@@ -0,0 +1,172 @@
...
+        validationResult.errors.push(ValidationErrors.PlaylistExceeds25Hours);
+      }
+
+      if (this.config.prolongPlaylistTo24Hours) {
+        //Informative warning, that playlist is under 24 hours and will be prolonged with placeholder videos
+        if (diffInHours < 24) {
+          validationResult.warnings.push(ValidationErrors.PlaylistProlongation);
+        }
+
+        //determine if playlist start date is older than 24 hours in the past
```
Zehelein: Should this rather be a general check not just for 24 hour prolonging case? Like if the playlist end is in the past give an error?

services/vod-to-live/service/src/domains/validation/playlist-published-validation-handler.ts
```diff
@@ -0,0 +1,172 @@
...
+          placeholderVideo,
+        ]);
+        if (mutualStreams.length < 1) {
+          validationResult.errors.push(
+            ValidationErrors.PlaylistVideosHaveNoMutualStreams,
+          );
+        }
+      } else {
+        validationResult.errors.push(
+          ValidationErrors.PlaylistVideosHaveNoMutualStreams,
```
Zehelein: That should be another error about missing placeholder or?

services/vod-to-live/service/src/domains/virtual-channel/virtual-channel-api.ts
```diff
@@ -0,0 +1,298 @@
...
+          headers: this.getAuthorizationHeader(),
+        },
+      );
+      return result.data;
+    } catch (error) {
+      throw getVirtualChannelApiMappedError(error);
+    }
+  };
+
+  public getPlaylistTransitions = async (
```
Zehelein, line 238: Please add a doc-string also to this/other public functions

services/vod-to-live/service/src/domains/smil/generator/playlist-smil-generator.ts
IrinaBazyleva: I understand the importance of maintaining consistency in the code, but I would like to request that this particular change be handled separately. The current task already involves modifying a large number of files, and adding this change on top of it may result in a less clean or organized pull request.

libs/media-messages/schemas/payloads/vod-to-live/commands/ensure-channel-live-command.json
IrinaBazyleva: The new naming applied with recent commit:
OLD NAME -> NEW NAME
`EnsureChannelLiveCommand` -> `CheckChannelJobStatusCommand`
`EnsureChannelLiveFailedEvent` -> `CheckChannelJobStatusFailedEvent`
`EnsureChannelLiveReadyEvent` -> `CheckChannelJobStatusSucceededEvent`

The `CheckChannelJobStatusCommand` is sent and received by the VOD-to-Live Service. On command receival a GET request is sent to USP Management API to check the status of a virtual channel. Once the channel is ready and the live stream has been successfully created, an event `CheckChannelJobStatusSucceededEvent` is sent with links to the stream. The Catalog API consumes these events.

The Check Channel Job Status is best that comes to my mind to describe the messages. That being said, I welcome any better proposals or alternatives.

services/vod-to-live/service/src/domains/cpix/generator/generate-cpix-settings.ts
IrinaBazyleva: The function was split into 2 new functions: `createDecryptionCpix` and `createEncryptionCpix`. Code was adjusted to use new functions.

services/vod-to-live/service/src/domains/live-stream/utils.ts
```diff
@@ -34,9 +34,9 @@ export const generatePlaylistFilePath = (
 };
 
 /**
- * Checks if provided string date-time is in the future.
- * @param date - string containing date time.
- * @returns - true if the provided date is in future.
+ * Checks whether the provided date-time string is in the future.e.
```
Zehelein: minor: ".e" is added to the end

services/vod-to-live/service/scripts/setup-webhook.ts
```diff
@@ -0,0 +1,123 @@
...
+      './scripts/resources/dev-user.json',
+      config.devServiceAccountClientId,
+      config.devServiceAccountClientSecret,
+      config.idServiceAuthBaseUrl,
+    );
+    const endpointUrl = urljoin(config.devChannelServiceBaseUrl, 'graphql');
+    const result = await axios.post(
+      endpointUrl,
+      {
+        query: print(gql`
```
AxTrusov, line 82: For me, this script hangs for a long time on the graphql request. If I run it explicitly via the graphiql UI - the same hanging happens. But I can run individual mutations separately and they will pass quickly. Strangely enough, the same does not happen with a similar entitlement script. Can you re-check?

services/vod-to-live/service/README.md
```diff
@@ -0,0 +1,250 @@
...
+
+`AZURE_BLOB_CONTAINER_NAME` - This is the mandatory setting. Name of the
+container in Azure Storage where the service will store channel data.
+
+#### DRM settings
+
+To enable DRM protection of the live stream, fill out the configurations below
+with the values available on the Axinom Portal.
+
+`KEY_SERVICE_API_BASE_URL` - This setting is optional.
```
AxTrusov, line 244: Technically, this is correct, but during development the variable must be commented out to truly be optional, as it will resolve to an empty string and will throw an error on startup if not specified. This is not specific to this case and is more of an env-var issue, which I have described here: https://github.com/evanshortiss/env-var/issues/167
Not sure if there is any decent workaround available for now, in case we want to disable DRM

services/vod-to-live/service/src/index.ts
```diff
@@ -0,0 +1,101 @@
...
+
+// Entry point for the service. For annotated version please see /services/media/service/src/index.ts.
+async function bootstrap(): Promise<void> {
+  handleGlobalErrors(logger);
+  setupGlobalSkipMaskMiddleware(trimErrorsSkipMaskMiddleware);
+  setupGlobalConsoleOverride(logger);
+  const app = express();
+  app.use(
+    bodyParser.json({
+      limit: '50mb',
```
AxTrusov: Minor thought: Would it make sense to make this value configurable?

services/vod-to-live/service/src/index.ts
```diff
@@ -0,0 +1,101 @@
...
+    config.azureBlobContainerName,
+  );
+
+  const virtualChannelApi = new VirtualChannelApi(
+    config.virtualChannelManagementApiBaseUrl,
+    config.virtualChannelManagementApiKey,
+  );
+
+  const keyServiceApi = new KeyServiceApi(config);
+  const broker = await setupMessagingBroker({
```
AxTrusov, line 57: This service has no database, so we cannot use the postgres counter we use in other services to support consistent message re-delivery policy. @Zehelein any thoughts?

services/vod-to-live/service/package.json
```diff
@@ -0,0 +1,58 @@
...
+    "env-var": "^6.3.0",
+    "express": "^4.18.1",
+    "media-messages": "^0.0.1",
+    "rascal": "^14.4.5",
+    "xmlbuilder2": "^3.0.2",
+    "url-join": "^4.0.1"
+  },
+  "devDependencies": {
+    "@types/jest": "^29",
+    "@types/pg": "^8.6.5",
```
AxTrusov: Since this service does not connect to the database, this dependency can probably be dropped?

services/vod-to-live/service/README.md
AxTrusov: Starting with version `0.32.0-rc.2`, `@axinom/mosaic-service-common` now includes an `asOptionalUrlString` `env-var` extension function. Please apply it to this variable.

services/vod-to-live/service/src/domains/validation/setup-pre-publishing-validation-webhook.ts
```diff
@@ -0,0 +1,87 @@
...
+    async (
+      req: Request<
+        Dict<string>,
+        Dict<string>,
+        WebhookRequestMessage<PrePublishingPayload>
+      >,
+      res: Response,
+    ) => {
+      const requestBody = req.body;
+      let validationResponse: WebhookResponse<PrePublishingPayload> | null =
```
AxTrusov: minor: can be set to just `let validationResponse: WebhookResponse<PrePublishingPayload>;`

services/vod-to-live/service/src/domains/validation/setup-pre-publishing-validation-webhook.ts
```diff
@@ -0,0 +1,87 @@
...
+        });
+        // generating response to the caller
+        validationResponse = generateWebhookResponse<PrePublishingPayload>({
+          payload: null,
+          errors: [ValidationErrors.ValidationNotPossible],
+          warnings: [],
+        });
+      }
+      //returns status code 400, if validation result has any errors, otherwise return 200
+      const statusCode = validationResponse.errors.length > 0 ? 400 : 200;
```
AxTrusov, line 82: Up for discussion: The way I see it, if an error is thrown for whatever reason - we need to generate a fitting response and send it with code 400+ (I guess 400 Bad Request is ok).

On the other hand, if no explicit error was thrown, and the validation handler returned some errors inside of the payload - we should consider it as a valid response (with errors) and return 200? right now it would respond with 400, which I'm not certan is correct...

services/vod-to-live/service/src/common/config/config-definitions.ts
```diff
@@ -0,0 +1,143 @@
...
+  const basicCustomizableConfigs = pick(
+    getBasicCustomizableConfigDefinitions(variables),
+    'tenantId',
+    'environmentId',
+    'idServiceAuthBaseUrl',
+  );
+  return {
+    ...getBasicConfigDefinitions(variables),
+    ...getBasicMetricsEndpointDefinitions(variables),
+    ...getBasicGraphQlConfigDefinitions(10200, variables),
```
AxTrusov: Default port should be 11900?

services/vod-to-live/service/src/common/config/config-definitions.ts
```diff
@@ -0,0 +1,143 @@
...
+  return getValidatedConfig(getConfigDefinitions(variables));
+};
+
+const config = getConfigType(getConfigDefinitions());
+/**
+ * The full Configuration type
+ */
+export type Config = typeof config;
+
+const dbConfig = getConfigType(getBasicDbConfigDefinitions());
```
AxTrusov: db-related parts can be dropped

services/vod-to-live/service/src/common/errors/validation-errors.ts
```diff
@@ -0,0 +1,50 @@
...
+  // General validation messages
+  ValidationFailed: {
+    message: `Unexpected error happened while validating message.`,
+    code: `VALIDATION_FAILED`,
+  },
+  ValidationNotPossible: {
+    message: `The message received cannot be validated.`,
+    code: `VALIDATION_NOT_POSSIBLE`,
+  },
+};
```
AxTrusov: Please add `as const` at the end `};` -> `} as const;` for a better IDE support:
Without:
image

With:
image

services/vod-to-live/service/src/domains/validation/utils.ts
```diff
@@ -0,0 +1,67 @@
...
+  video: DetailedVideo,
+  config: Config,
+): ValidationResult => {
+  const validationResult: ValidationResult = {
+    errors: [],
+    warnings: [],
+  };
+  if (video.video_encoding.is_protected) {
+    if (config.isDrmEnabled) {
+      if (video.video_encoding.video_streams.find((s) => !s.key_id)) {
```
AxTrusov: Please adjust this not to include Subtitle and Caption streams, they always have null key IDs. And cover this case with unit test(s).

services/vod-to-live/service/src/domains/validation/utils.ts
```diff
@@ -0,0 +1,67 @@
...
+  if (video.video_encoding.output_format !== 'CMAF') {
+    validationResult.errors.push({
+      message: `Video ${video.id} output format is '${video.video_encoding.output_format}'. Expected output format 'CMAF'`,
+      code: 'WRONG_VIDEO_FORMAT',
+    });
+  }
+
+  if (!video.video_encoding.video_streams.find((s) => s.type === 'AUDIO')) {
+    validationResult.errors.push({
+      message: `Video ${video.id} is missing AUDIO stream.`,
```
AxTrusov: `Video ${video.id} is missing AUDIO stream(s).`

services/vod-to-live/service/src/domains/smil/models/utils.ts
```diff
@@ -0,0 +1,132 @@
...
+): HeaderMetadata[] => {
+  return [
+    createHeaderMetadata(HeaderMetadataNames.Vod2Live, true),
+    createHeaderMetadata(
+      HeaderMetadataNames.Vod2LiveStartTime,
+      vod2LiveStartTime.toISOString(),
+    ),
+    //`splice_media` should be set to `true`
+    // but is set to `false` temporary due to bug in Origin affecting the HLS streams
+    //todo: set `splice_media` to `true`, when Origin bug is fixed
```
AxTrusov: Would be nice to also add a reference to the bug, e.g. a link to a github issue or similar. If it exists ofc

services/vod-to-live/service/src/domains/cpix/models/cpix-request.ts
```diff
@@ -0,0 +1,94 @@
...
+  '@kid': string;
+  '@commonEncryptionScheme': CommonEncryptionScheme;
+}
+
+export interface ContentKeyUsageRuleList {
+  'cpix:ContentKeyUsageRule': ContentKeyUsageRule[];
+}
+
+export interface ContentKeyUsageRule {
+  'cpix:VideoFilter': any;
```
AxTrusov: I think all `any` occasions in this file can be safely replaced with `unknown`. If not - I would add an eslint comment at the top of this file, e.g. `/* eslint-disable @typescript-eslint/no-explicit-any */`

AxTrusov: same for utils.ts

services/vod-to-live/service/src/domains/utils.ts
```diff
@@ -0,0 +1,41 @@
...
+import { SECOND_IN_MILLISECONDS } from '../common';
+
+/**
+ * Converts any object to string XML representation.
+ * @param expandedObject - object to convert to XML.
+ * @returns- string containing XML of the object.
+ */
+export const convertObjectToXml = (
+  expandedObject:
+    | {
```
AxTrusov: I would replace this type with `string | ExpandObject`, unless an import like `import { ExpandObject } from 'xmlbuilder2/lib/interfaces';` breaks anything...

services/vod-to-live/service/.env.template
```diff
@@ -0,0 +1,45 @@
...
+# Live stream DRM protection
+KEY_SERVICE_API_BASE_URL=
+KEY_SERVICE_TENANT_ID=
+KEY_SERVICE_MANAGEMENT_KEY=
+DRM_KEY_SEED_ID=
+
+# Feature flag. When turned on every playlist with duration under 24 hours
+# will be automatically prolonged to hit 24 hour duration.
+# Defaults to TRUE
+PROLONG_PLAYLIST_TO_24_HOURS=TRUE
```
AxTrusov: @Zehelein I would either change the default to FALSE, or add a comment to Strongly recommend setting the value to FALSE during development. When it is set to true, Virtual Channel would take an eternity to generate a 24h long livestream from the playlist, so you pretty much have to set it to false to be able to test.

services/vod-to-live/service/src/messaging/handlers/check-channel-job-status-handler.ts
```diff
@@ -0,0 +1,125 @@
...
+        );
+
+      this.logger.error({
+        message: `Channel ${channel_id} failed to go live.`,
+        details: {
+          uspErrors,
+        },
+      });
+
+      await this.broker.publish<CheckChannelJobStatusFailedEvent>(
```
AxTrusov, line 82: Should some service receive this (CheckChannelJobStatusFailedEvent) message and act accordingly?

services/vod-to-live/service/src/domains/utils.ts
AxTrusov, line 1: minor: I'd move this file and models.ts into either src/common folder or src/domains/common

services/vod-to-live/service/.env.template
Zehelein: Ok yes - likely best to default it to false. During development "nobody" will fill 24 hours and for production they should anyways try to manually fill 24 hours (for a 24/7 live stream).

services/vod-to-live/service/src/domains/validation/setup-pre-publishing-validation-webhook.ts
Zehelein: What is our general pattern here? I suggest returning a 400 error code in case of errors. So we should IMO instead adjust the other places. For GraphQL this is of course another story - there the specification dictates a 200 response with an error object. But for REST-based webhooks, we should consider the HTTP codes IMO.

AxTrusov: We don't have enough webhooks to have a general pattern yet (and the entitlement ones are not producing custom validation messages). I was just debating if validation error is equal to thrown error. As long as returned payload errors are correctly visualized and processed - I guess it should be fine.

services/vod-to-live/service/src/domains/smil/models/utils.ts
AxTrusov: The bug is internal to the origin components, so there is no public link to it. Adjusted the comment with a bit more details.

services/vod-to-live/service/src/index.ts
AxTrusov: Discussed, acceptable for now. Adding a database just for this case is too much.

services/vod-to-live/service/src/messaging/handlers/check-channel-job-status-handler.ts
AxTrusov: Just logging the error is acceptable for now (in the catalog service). Extra handling should be considered as part of custom adjustments.

services/vod-to-live/service/src/common/config/config-definitions.ts
```diff
@@ -117,6 +118,13 @@ export const getConfigDefinitions = (
         .get('CHANNEL_PROCESSING_WAIT_TIME_IN_SECONDS')
         .default('600')
         .asIntPositive(),
+
+    /**
+     * Sets the body size limit for the webhook requests, since playlists can be
+     * quite large. Default is 50mb. Examples: https://github.com/visionmedia/bytes.js#bytesparsestringnumber-value-numbernull
+     */
+    webhookBodySizeLimit: () =>
+      env.get('WEBHOOK_BODY_SIZE_LIMIT').default('50mb').asString(),
```
Zehelein, line 127: How is this guaranteed to work? The "bytes" library that is linked is not a dependency of the "env-var" package (according to the npm website).

AxTrusov: The `bytes` library is used internally by `json` component of the body-parser library used by the express app. 
It also supports string/numeric values (as seen in the example link), so I did not see a reason to add hard restrictions on value ending with mb, kb, etc..

services/vod-to-live/service/src/domains/smil/generator/playlist-smil-generator.ts
AxTrusov: Separate work item 40235 created

AxTrusov (approved):

PR #80, opened 2023-06-05
[AB#39600] [AB#39414] [AB#39279] [AB#39062] Movies and movie genres data can now be synced with the Localization service

AxTrusov:
# Pull Request

## Prerequisites

- [x] All tests pass

## Description

Genral description:
-

- Integration with the Localization service is added for Movies and Movie Genres.
- By enabling the Logical Replication on the media Service database - it is now possible to listen to changes associated with localizable tables and send out messages to the Localization Service.
	- Check `infra/docker-compose.yml` for changes that are required to be done on the Postgres Server level to enable Logical Replication
	- Owner user also must be altered to include Replication (e.g. see `scripts/helpers/db-command-helpers.ts` for sample script.
- To account for existing movies and movie genres - a one-time script is added to sync existing movies with the localization service `src/one-time-register-movie-localization-sources.ts`.
    - It's part of the running service and must be launched from within the deployment environment (e.g. using `node dist/one-time-register-movie-localization-sources`, which assumes that process.env already has config values loaded)
- A dedicated `IS_LOCALIZATION_ENABLED` config variable is added to enable/disable localization functionality in the service (does not affect the database migrations/setup, but affects the data syncing between media and localization services)
- service account token is now cached in memory (to avoid spamming its retrieval, e.g. in case of syncing movies data with localization service during ingest)
- Movie and Movie Genre localizable entity definitions are now synced on media service startup.
- Media Workflows adjusted to connect movie genres and movie details with localization service workflows.

Testing notes:
- 

- Following actions trigger messages to the localization service, the results of which can be observed in localization stations. Make sure that at least one non-default locale is created for the environment for changes to be observable.
	- Create a movie - movie localization created
	- Update movie title and/or description and/or synopsis - movie localization updated
	- Assign/Change/Unassign movie cover (not teaser) - movie localization updated
	- Delete movie - movie localization deleted
	- Create a movie genre -  movie genre localization created
	- Update movie genre title - movie genre localization updated
	- Delete movie genre - movie genre localization deleted
- Ingest of movies must be tested as well


services/media/service/src/domains/common/utils/localization-utils.ts
```diff
@@ -0,0 +1,79 @@
...
+};
+
+export const assertLocalizationData: (
+  data: Dict<unknown> | undefined,
+  tableName: Table,
+) => asserts data is Dict<unknown> = (
+  data: Dict<unknown> | undefined,
+  tableName: Table,
+): asserts data is Dict<unknown> => {
+  if (!data) {
```
Zehelein, line 60: To be really sure (as it is likely that it can come from a message that is not properly asserted): maybe add `typeof data !== 'object'`.

Maybe we can then rather generalize it as "assertObject" and further down "assertObjectField" or such?

services/media/service/src/domains/common/utils/localization-utils.ts
```diff
@@ -0,0 +1,79 @@
...
+    });
+  }
+};
+
+export const assertLocalizationColumn = (
+  data: Dict<unknown>,
+  columnName: string,
+  tableName: Table,
+): void => {
+  if (!(columnName in data)) {
```
Zehelein, line 73: Maybe also here to be really sure: `typeof obj !== 'object' || obj === null `

services/media/service/src/domains/movies/localization/movies-replication-handlers.ts
```diff
@@ -0,0 +1,85 @@
...
+  const entityType = LOCALIZATION_MOVIE_TYPE;
+  const fieldDefinitions = MovieFieldDefinitions.filter((d) => !d.is_archived);
+  return {
+    insertHandler: async (newData: Dict<unknown> | undefined) => {
+      assertMovie(newData);
+
+      const fields: Dict<unknown> = {};
+      for (const { field_name } of fieldDefinitions) {
+        const value = newData[field_name];
+        if (value) {
```
Zehelein: An if-check in this way is likely not right. It checks for all null-ish values and empty string vs. undefined/null makes a difference that should also be sent. As for the movie genres handler such a check is also not there.

services/media/service/src/domains/movies/localization/movies-replication-handlers.ts
```diff
@@ -0,0 +1,85 @@
...
+      );
+    },
+    updateHandler: async (
+      newData: Dict<unknown> | undefined,
+      oldData: Dict<unknown> | undefined,
+    ) => {
+      assertMovie(newData);
+      assertMovie(oldData);
+
+      const fields: Dict<unknown> = {};
```
Zehelein: This construct is used in multiple places - also the above one for the "insertHandler". Maybe good to extract it into a reusable function (especially as episodes etc. will follow later).

services/media/service/src/domains/movies/localization/movies-replication-handlers.ts
AxTrusov: - This would be true for `updateHandler`, and there such check is not present,
  so whatever is set - we send out as long as data was changed.
- When we insert the movies, we only care about sending the data to the localization
  service only for fields with actual values (e.g. just title, or title +
  description + synopsis), because empty values would any way be defaulted to empty in localization service. This makes the sent message a bit smaller. This might look like a tiny optimization on the surface, but considering scenarios of ingesting thousands of movies/episodes - it would add up.
  -  I also did not use isNullOrWhitespace for the check because title is guaranteed to have a non-empty value due to db constraint, and for description/synopsis empty strings are acceptable in the current implemenation
- For movie genres such a check is not needed, as movie genres only have the
  required title field (as opposed to movies that also have synopsis/description).
  - But in context of generalizing the logic the check is still acceptable

services/media/service/src/domains/common/utils/localization-utils.ts
AxTrusov: This data would come from logical replication service, so the responsibility
falls to the PgoutputPlugin to produce an object, which so far produced only null, undefined, or
object values (depending on operation type and db setup). Also, the thrown errors are localization-specific, so I did not
want to generalize it even more.

I think the current implementation should be good enough and would only require manual checks when
related npm packages are updated (e.g. PgOutputPlugin gets an update and somehow
starts producing data of non-object types), but if you think it would not hurt to check
for typeof 'object' anyway - then I would separate the check into a separate
`if` block and throw a more specific error in that case.

services/media/service/src/domains/common/utils/localization-utils.ts
AxTrusov: The reply above is slightly related to this one.
If we stick to not generalizing this function, then there is no need to check
this, because `assertLocalizationColumn` and `assertLocalizationData` are meant
to be checked in combination with one another to be able to produce
understandable localization-specific errors that are matching the failure condition.

services/media/service/src/common/utils/token-utils.ts
```diff
@@ -1,23 +1,33 @@
 import {
   generateLongLivedToken,
   getServiceAccountToken,
-  TokenResult,
 } from '@axinom/mosaic-id-link-be';
+import NodeCache from 'node-cache';
```
Zehelein, line 5: Is there/can we have a check if we get an authorization error that we try (once) to get a new token? Like if the secret key was regenerated that we do not have to wait for the cache to expire.

Zehelein (approved):
Added one comment but looks good otherwise - thanks :-)

AxTrusov:
This PR is being put on hold for the time being. Things to consider when it is resumed:
- sync with `dev` branch and make sure everything still works, do any related adjustments if necessary.
- `UpsertLocalizationSourceEntityFields` - these types would no longer be generated, so need to make adjustments to replace them with `Record<string,unknonw>`
- Resolve Frank's comments. (one of them already has a reply, probably just needs to be resolved without actions)
- Adjust the default value for the logical replication slot name. Make it empty, adjust the description comment to signify its importance and uniqueness, adjust the setup script to generate the value based on some other config values (e.g. envirnoment_id and service_id), and maybe provide a helper script to generate the actual value for deployed environments, need to confirm if this needs to be a `setup` script or a separate custom helper script.
   - This was moved to a separate task #40804
- Make sure that hosting service is properly handling logical replication slots

services/media/service/src/common/utils/token-utils.ts
AxTrusov: There is no good way for now, but we already discussed this with Ruwan and came up with an appropriate solution which requires extra changes in ID libs. Related work item is #40610

PR #98, opened 2023-06-26
[AB#40278] ajv deduplication and mosaic depdencies update

AxTrusov:
# Pull Request

## Prerequisites

- [x] All tests pass

## Description

- ajv dependency deduplication to fix build issues with updated @axinom/mosaic-cli
- mosaic dependencies updated to latest pre-release version
- updated message JSON schemas to work with latest asyncapi tooling


libs/media-messages/src/generated/schemas/payloads/media/commands/check-finish-ingest-item-command.json
```diff
@@ -10,10 +10,10 @@
   ],
   "properties": {
     "ingest_item_step_id": {
-      "description": "A unique identifier that is used to track ingest messages for a single document to keep ingest document status up-to-date.",
       "type": "string",
       "minLength": 1,
-      "pattern": "^$|.*\\S.*"
+      "pattern": "^$|.*\\S.*",
+      "description": "String of minimum length of 1 character, which is also cannot consist of only whitespace characters."
```
Zehelein: A string with at least one character and not only whitespace characters.

libs/media-messages/src/generated/types/payloads/media/commands/check-finish-ingest-item-command.ts
```diff
@@ -3,7 +3,7 @@
  */
 export interface CheckFinishIngestItemCommand {
   /**
-   * A unique identifier that is used to track ingest messages for a single document to keep ingest document status up-to-date.
+   * String of minimum length of 1 character, which is also cannot consist of only whitespace characters.
```
Zehelein: Here and further down we should keep the original message as it is much more helpful.

libs/media-messages/src/generated/types/payloads/media/commands/delete-entity-command.ts
```diff
@@ -19,8 +18,5 @@ export interface DeleteEntityCommand {
    * The name of the entity type.
    */
   entity_type?: string;
-  /**
-   * Additional input data.
-   */
-  input?: BulkEntityInput;
+  input?: { [name: string]: unknown };
```
Zehelein, line 21: The description got lost here and with the externalized type (and the description there) it was much clearer what this is used for.

EmperorRXF (approved):
@Zehelein, I just had a chat with @AxTrusov about this PR now. Think we both agree that the "documentation breaking change" is not ideal. At the same time, the code changes this PR is applying on media-template is quite important, as without them we leave the template in a broken state when do release the latest version.

So we're left with two choices rn, revert the libs changes we did on core-libs or go forward with this PR on media-template.

I now had a look at the original reason for us to bump the codegen libs, and it was due to a critical security alert we got via dependabot for possible code-injections (https://github.com/Axinom/Navy/security/dependabot/88)

So, considering all of this, think we can go forward with this PR - knowing that we have a follow-up task to address the breaking documentation change (soonish..). In the meantime, we can expect OSS users to find existing usages of the generated types already in code, and to some-degree derive the purpose of the individual fields of the types. In worse case, we can provide needed support if/when they reach out to us.

libs/media-messages/src/generated/types/payloads/media/commands/check-finish-ingest-item-command.ts
AxTrusov: Related to Ruwans comment: https://github.com/Axinom/mosaic-media-template/pull/98#pullrequestreview-1500138990

libs/media-messages/src/generated/types/payloads/media/commands/delete-entity-command.ts
AxTrusov: Related to Ruwans comment: https://github.com/Axinom/mosaic-media-template/pull/98#pullrequestreview-1500138990

PR #102, opened 2023-07-05
[AB#40100] feat: build and run container as their native non-root users

yohanAnushka:
# Pull Request

## Prerequisites

- [x] All tests pass

## Description

- Backend Service Containers (node):
   - ownership of `/app` directory is given to `node` user that is built into the standard `node:18-alpine` image and has `1000` as the `uid` and `gid`. This change was not needed for most of the services but if the app itself or a library tries to download something like geolite data, there will be a permission issue so now the ownership is given to `node` user universally
   - container is run as user `node(1000)` and group `node(1000)` with no privilege escalations


abrykovskii (approved):
LGTM

PR #106, opened 2023-07-13
[AB#40442] feat: set remote fspg instance creds in env file

yohanAnushka:
# Pull Request

## Prerequisites

- [x] All tests pass

## Description

- Running lengthy tests that involve Postgres databases consume a lot of memory and sometimes Microsoft build agents are not able to accommodate all the memory requests. This leads to builds crashing.
- Utilize existing non-production FSPG instance as the database source so that we can skip running tests against a Postgres container inside the built agent, completely skipping the memory penalty.

### DIFF
--- a/scripts/helpers/db-command-helpers.ts
+++ b/scripts/helpers/db-command-helpers.ts
@@ -85,13 +85,10 @@ export const runResetQueries = async (
   dbLoginPassword: string,
   dbOwner: string,
   dbOwnerPassword: string,
-  dbEnvOwner?: string,
-  dbEnvOwnerPassword?: string,
   pgRoot?: string,
 ): Promise<void> => {
   const client = await pgPool.connect();
   try {
-    const createEnvOwner = dbEnvOwner && dbEnvOwnerPassword;
     const commands: string[] = [
       // Drop existing connections to database
       `SELECT pg_terminate_backend(pid) FROM pg_stat_activity WHERE datname='${dbName}';`,
@@ -102,18 +99,12 @@ export const runResetQueries = async (
       `DROP ROLE IF EXISTS ${dbGqlRole};`,
       `DROP ROLE IF EXISTS ${dbLogin};`,
       `DROP ROLE IF EXISTS ${dbOwner};`,
-      createEnvOwner ? `DROP ROLE IF EXISTS ${dbEnvOwner};` : '',
 
       // Create a non-superuser user (unlike PostGraphile suggests) because we want to make the development database behave as close to the deployment database as possible. E.g. when deploying to Azure or AWS owner user will never be a superuser.
       `CREATE ROLE ${dbOwner} WITH LOGIN PASSWORD '${dbOwnerPassword}';`,
 
       // This is the no-access role that PostGraphile will run as by default
       `CREATE ROLE ${dbLogin} WITH LOGIN PASSWORD '${dbLoginPassword}' NOINHERIT;`,
-
-      // A second role similar to owner but without the permission to bypass row level security
-      createEnvOwner
-        ? `CREATE ROLE ${dbEnvOwner} WITH LOGIN PASSWORD '${dbEnvOwnerPassword}' NOINHERIT;`
-        : '',
 
       // This is the role that PostGraphile will switch to (from ${DATABASE_LOGIN}) during a GraphQL request
       // Basically, this is a group role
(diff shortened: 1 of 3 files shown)


scripts/helpers/db-command-helpers.ts
```diff
@@ -126,6 +127,8 @@ export const runResetQueries = async (
       // When owner is superuser - this is not needed, but in deployed environments and for test databases owner is a regular user and also needs such grant.
       `GRANT ${dbGqlRole} TO ${dbOwner};`,
       createEnvOwner ? `GRANT ${dbGqlRole} TO ${dbEnvOwner};` : '',
+      // By granting owner_role to the root role, we can then create the database while settin owner_role as the owner.
+      pgRoot ? `GRANT ${dbOwner} TO ${pgRoot}` : '',
```
AxTrusov, line 124: This seems strange. Shouldn't the root user have access to the whole postgres server anyway?

.env.template
```diff
@@ -31,16 +31,16 @@ POSTGRESQL_HOST=localhost
...
+POSTGRESQL_ROOT=root
+POSTGRESQL_ROOT_PASSWORD=root
 
 # pgAdmin
 PGADMIN_PORT=9999
 
 # RabbitMQ Connection
 RABBITMQ_PROTOCOL=amqps
-RABBITMQ_HOST=rabbitmq.eu.axinom.net
+RABBITMQ_HOST=rabbitmq.test.axinom.net
```
AxTrusov: The template should ideally not be changed. we should continue using prod urls and I would prefer to keep old credentials, otherwise, there is probably a need to fully reset dev postgres server...

yohanAnushka: @AxTrusov Why are we running tests against prod?

scripts/helpers/db-command-helpers.ts
yohanAnushka: In FSPGs that's not necessarily the case and the database creation would fail without this grant.

abrykovskii (approved):
LGTM

scripts/helpers/db-test-helpers.ts
```diff
@@ -65,6 +65,9 @@ export async function recreateTestDbTemplate(
     dbConfig.dbLoginPassword,
     dbConfig.dbOwner,
     dbConfig.dbOwnerPassword,
+    undefined,
```
AxTrusov: Since env owner main use case is in multitenant services and atm this is not used in media template - please drop `dbEnvOwner` and `dbEnvOwnerPassword` parameters from the `runResetQueries` function. If there is ever a need to introduce such roles to some media template service - it can be done at that point in time

scripts/helpers/db-command-helpers.ts
```diff
@@ -126,6 +127,8 @@ export const runResetQueries = async (
       // When owner is superuser - this is not needed, but in deployed environments and for test databases owner is a regular user and also needs such grant.
       `GRANT ${dbGqlRole} TO ${dbOwner};`,
       createEnvOwner ? `GRANT ${dbGqlRole} TO ${dbEnvOwner};` : '',
+      // By granting owner_role to the root role, we can then create the database while settin owner_role as the owner.
```
AxTrusov: I would change this to something that explains why the root user even needs such a grant in the first place. e.g. `In some systems (e.g. --FSPG--), it is not possible to get the actual ROOT user credentials, so a user with a lot of privileges is used instead and we need to grant db owner rights to it.`

`--FSPG--` should be replaced with something understandable for regular users, even I have not found what it stands for. Maybe replace with `hosted Azure or AWE database servers` or something technically-correct along those lines.

.env.template
```diff
@@ -31,16 +31,16 @@ POSTGRESQL_HOST=localhost
 POSTGRESQL_PORT=5432
 POSTGRESQL_USER_SUFFIX=
 PGSSLMODE=disable
-POSTGRESQL_ROOT=postgres
-POSTGRESQL_ROOT_PASSWORD=postgres
+POSTGRESQL_ROOT=root
```
AxTrusov: As already mentioned in another comment below, please revert the credentials values, Such change is only needed for PR builds and would require developers to re-create their local databases without an explicit need. This should be reflected in PR/Build-specific files.

.env.template
AxTrusov: we do not, the message-sending parts are mocked in tests

AxTrusov (approved):

PR #115, opened 2023-07-24
[AB#39602] channel localizations added

AxTrusov:
# Pull Request

## Prerequisites

- [x] All tests pass

## Testing Notes
- Publish a channel with localizations enabled or disabled - both cases would produce at least one localization (either for only default locale, or for default locale and non-default locales)
- Request a channel with specific localization to get the values for that specific localization
    - If requested without adding a locale filter - default localization values are returned
    - if non-existent locale is specified in the locale filter - default localization values are returned.

## Additional Notes
- when `@axinom/mosaic-messages` is updated - this PR should be adjusted to use more concrete types (currently using `any` in a few spots)
- Generalized plugin added to allow adding `localization` subobject relation.
- Existing title/description columns from channel are migrated into the dedicated localizations table


services/catalog/service/migrations/committed/000011-channel-localizations-added.sql
```diff
@@ -0,0 +1,31 @@
+--! Previous: sha1:c9457b917399b9a43ac0fc04d0e130cf728bfcb7
+--! Hash: sha1:7afd39916d603525b3b67b5a836d90b445e8bf2b
+--! Message: channel-localizations-added
+
+CREATE TABLE IF NOT EXISTS app_public.channel_localizations(
+  id INTEGER PRIMARY KEY GENERATED BY DEFAULT AS IDENTITY,
+  channel_id TEXT REFERENCES channel ON DELETE CASCADE,
+  locale TEXT NOT NULL,
+  is_default_locale BOOLEAN NOT NULL,
+  title TEXT,
```
Zehelein: The title should be a required property or? We should then mark the field as "not null".

services/catalog/service/src/domains/channels/plugins/all-channel-plugins.ts
```diff
@@ -0,0 +1,6 @@
+import { makePluginByCombiningPlugins } from 'graphile-utils';
+import { AssetLocalizationPlugin } from '../../../graphql/plugins';
+
+export const AllChannelPlugins = makePluginByCombiningPlugins(
+  AssetLocalizationPlugin('channel', 'channel_localizations'),
```
Zehelein: As mentioned: please change to something like "EntityLocalizationPlugin"

services/catalog/service/src/graphql/plugins/asset-localization-plugins.ts
```diff
@@ -0,0 +1,113 @@
...
+  gql,
+  makeExtendSchemaPlugin,
+  makePluginByCombiningPlugins,
+} from 'graphile-utils';
+
+import { makeJSONPgSmartTagsPlugin } from 'graphile-utils';
+import { selectOne } from 'zapatos/db';
+import { Table } from 'zapatos/schema';
+
+const AssetLocalizationSmartTagsPlugin = (assetName: string): Plugin =>
```
Zehelein: Asset --> entity in all places

services/catalog/service/src/graphql/plugins/asset-localization-plugins.ts
```diff
@@ -0,0 +1,113 @@
...
+    const capitalizedAssetName = inflection.upperCamelCase(assetName);
+    return {
+      typeDefs: gql`
+        extend type ${capitalizedAssetName} {
+          """
+          The localized ${assetName} metadata.
+          """
+          localization(
+            """
+            The localization is returned for the specified locale. If not defined, the default locale will be used instead.
```
Zehelein: The localizable field values are returned for the requested locale. In case the locale is not provided or does not exist, the default locale will be used.

services/catalog/service/src/graphql/plugins/asset-localization-plugins.ts
```diff
@@ -0,0 +1,113 @@
...
+            """
+            locale: String
+          ): ${capitalizedAssetName}Localization! @requires(columns: ["id"])
+        }
+      `,
+      resolvers: {
+        [capitalizedAssetName]: {
+          localization: async ({ id }: { id: string }, args, context) => {
+            try {
+              const localization = await selectOne(localizationTable, {
```
Zehelein: Also just for reference as we did talk about it: N+1 issue

services/catalog/service/src/tests/channels/localizations.db.spec.ts
```diff
@@ -0,0 +1,207 @@
...
+      is_default_locale: true,
+    }).run(ctx.ownerPool);
+  });
+
+  afterEach(async () => {
+    await ctx?.truncate('channel_localizations');
+  });
+
+  afterAll(async () => {
+    await ctx?.truncate('channel');
```
Zehelein, line 44: `?` either add it for both lines or none

services/catalog/service/src/tests/channels/localizations.db.spec.ts
```diff
@@ -0,0 +1,207 @@
...
+      {
+        localization: {
+          title: 'Default title',
+          description: 'Default description',
+        },
+      },
+    ]);
+  });
+
+  it('Channel with only 2 localizations and empty filter -> default localization returned', async () => {
```
Zehelein: why the "only"?

services/catalog/service/migrations/committed/000011-channel-localizations-added.sql
AxTrusov: Yes, will mark as not null. I was thinking from future perspective where there might be fallback options and some translations might not have titles, but for now making title required should be more correct and this can be adjusted in the future.

services/catalog/service/src/tests/channels/localizations.db.spec.ts
AxTrusov: copy-paste ghost

services/catalog/service/src/graphql/plugins/asset-localization-plugins.ts
AxTrusov: Adjusted. Sources:
- https://www.graphile.org/postgraphile/make-extend-schema-plugin/#using-the-pgquery-directive-for-non-root-queries-and-better-performance
- https://github.com/graphile/graphile-engine/blob/5211758b7a48191ffd7600f9f5ae572672ffd221/packages/graphile-utils/__tests__/ExtendSchemaPlugin-pg.test.js#L507-L720

services/catalog/service/src/graphql/plugins/entity-localization-plugins.ts
```diff
@@ -0,0 +1,105 @@
...
+                  sql.identifier(entityName);
+                const localizations_table = queryBuilder.getTableAlias();
+                const fk = sql.identifier(`${entityName}_id`);
+                const isDefault = sql.fragment`${localizations_table}.is_default_locale IS TRUE`;
+                const condition = args.locale
+                  ? sql.fragment`(${localizations_table}.locale = ${sql.value(
+                      args.locale,
+                    )} OR ${isDefault})`
+                  : isDefault;
+                queryBuilder.where(
```
AxTrusov, line 84: With the limitations of this approach, adjusted it to perform the following:
- Always include default localization into the query, but use of orderBy to make sure it is the last in the list
- apply `limit` to only take the first entry. 
   - If existing locale is specified - record for it will always be first and `limit 1` will remove the default locale from final result.
   - If invalid locale value is specified - only default locale is returned and it is taken

Zehelein: thanks - this sounds great

Zehelein (approved):

PR #128, opened 2023-08-02
[AB#39601] movie and movie genre localizations can now be published

AxTrusov:
[[AB#39601](https://dev.azure.com/axinom/5e61ad7f-f562-45f3-8078-9ade7df639c1/_workitems/edit/39601)]

# Pull Request

## Prerequisites

- [x] All tests pass

## Description
- Previous PRs included syncing source changes of movies and movie genres to the localization service.
- This PR now supports publishing of localizations from the localization service (through media service) into the catalog service.
- Similar to channel localization publishing, movie and movie genre GraphQL schema was adjusted to no longer have localizable properties on the root of the returned entity object, moving into the dedicated `localization` subobject.
   - this supports both cases when localization is enabled and disabled.

## Testing notes
- Two ways of publishing movies and movie genres should be checked - when IS_LOCALIZATION_ENABLED flag in the media service is set to true or false. 
   - When true - publish validation would include localization-related messages for incomplete localization and include `localizations` array for each published movie genre and movie
   - When false - publish validation should not include localization-related messages and `localizations` array should not be included in the published JSON
- Check the catalog API to make sure that retrieving localization without the local filter or with incorrect locale value would return default value(s) and with specific code would return value(s) for specific locale


services/catalog/service/src/domains/movies/handlers/movie-genres-unpublished-event-handler.db.spec.ts
```diff
@@ -27,28 +28,49 @@ describe('MovieGenrePublishEventHandler', () => {
       // Arrange
```
Zehelein, line 28: The test description could be improved: "Received a movie-genres unpublish message - the single existing movie-genre is removed". Reason: it is not like we receive a message to unpublish a single movie-genre but generally we receive a message to unpublish all movie genres always. Similar for the other test cases.

services/catalog/service/src/domains/movies/handlers/movie-genres-unpublished-event-handler.db.spec.ts
```diff
@@ -27,28 +28,49 @@ describe('MovieGenrePublishEventHandler', () => {
...
       const movieGenre = await select('movie_genre', all).run(ctx.ownerPool);
+      const localizations = await select('movie_genre_localizations', all).run(
+        ctx.ownerPool,
+      );
 
       expect(movieGenre).toHaveLength(0);
+      expect(localizations).toHaveLength(0);
     });
 
     test('two existing movie-genres are unpublished', async () => {
```
Zehelein: Pleas also add one (single) test to show unpublishing works with not just default locale but 1-2 others as well.

services/catalog/service/src/domains/movies/handlers/movie-unpublished-event-handler.db.spec.ts
```diff
@@ -26,9 +27,14 @@ describe('MoviePublishEventHandler', () => {
...
+      await insert('movie', { id: 'movie-1' }).run(ctx.ownerPool);
+      await insert('movie_localizations', {
+        movie_id: 'movie-1',
+        title: 'Some title',
+        description: 'testing',
+        locale: DEFAULT_LOCALE_TAG,
+        is_default_locale: true,
+      }).run(ctx.ownerPool);
 
       const message: MovieUnpublishedEvent = { content_id: 'movie-1' };
```
Zehelein, line 65: Please check also that images/videos/localizations/... don't exist anymore plus relations to genres and similar.

services/media/service/src/domains/common/utils/get-localizations-metadata.spec.ts
```diff
@@ -0,0 +1,264 @@
...
+  });
+
+  afterEach(async () => {
+    validationResult = () => undefined;
+  });
+  afterAll(async () => {
+    jest.restoreAllMocks();
+  });
+
+  it('Empty localizations array is returned -> empty localizations array', async () => {
```
Zehelein, line 76: Please add also a successful test that returns actual localized values.

services/media/service/src/domains/movies/handlers/publishing-movie-genres-processor.ts
```diff
@@ -1,40 +1,63 @@
...
   }).run(queryable);
 
-  const snapshotJson: MovieGenresPublishedEvent = {
-    genres: genres.map((genre) => ({
-      title: genre.title,
+  const validations: SnapshotValidationResult[] = [];
+  const mappedGenres: MovieGenre[] = [];
+  for (const genre of genres) {
+    const { result: localizations, validation: localizationsValidation } =
+      await getMovieGenreLocalizationsMetadata(
```
Zehelein, line 32: Could this be changed to send a single GQL query with like 20 genres at once to not have 20 separate requests?

AxTrusov: Not at the moment. Created a separate work item to optimize it in the future. #40874

PR #133, opened 2023-08-23
[AB#40224] service ID and urls are renamed from encoding to video

AxTrusov:
# Pull Request

## Prerequisites

- [x] All tests pass

## Description

Urls and service ID reference updated to use `video` instead of `encoding`.


services/entitlement/service/README.md
```diff
@@ -83,7 +83,7 @@ The `yarn setup:webhooks` script is used for development environments. To get
 the values for `ENTITLEMENT_WEBHOOK_SECRET` and `MANIFEST_WEBHOOK_SECRET` env
 variables for the production environment, the following mutation must be
 launched against the Video Service GraphQL API. You can use GraphiQL UI for that
-(https://encoding.service.eu.axinom.net/graphiql):
+(https://video.service.eu.axinom.net/graphiql):
```
AxTrusov, line 86: This URL is currently not accessible. Need to make sure it is before this PR is completed

services/media/service/src/generated/messaging/rascal-schema.json
```diff
@@ -407,7 +407,7 @@
         },
         "media-service:cue_point_types:declared": {
```
AxTrusov: Currently, a change like this would make sure that we receive messages using the updated service ID, but:
1. the previously existed queue` ax-encoding-service:cue_point_types:declare` in media-template-specific vhost will continue to exist.
2. The queue for events, e.g. `media-service:cue_point_types:declared` would have bindings for both service IDs: 
image
   a. This will result in media service receiving a message 2 times because of our backward compatibility logic.

In short - we need to perform some sort of RabbitMQ cleanup to remove obsolete queues and bindings

services/entitlement/service/README.md
Zehelein: Wrote with Anushka and he said the change is in the repo and once it is pushed the URL will work.

services/media/service/src/generated/messaging/rascal-schema.json
Zehelein: Yes right - the cleanup should be done also as part of this PR - thanks.

Zehelein: Was done as part of https://github.com/Axinom/mosaic-media-template/pull/140/files

Zehelein (approved):

PR #135, opened 2023-08-24
Update Media Template to use mosaic-grpahql-common and service-common after split

yasithA:
- Update media-template to use the new mosaic-graphql-common library and service-common libraries.
- Update unit tests.

TODO:
Re-check changes to the entitlement service. Might have to think of changes to the graphql-common lib instead of the changes here.


yasithA:
Closing this PR since this is handled through https://github.com/Axinom/mosaic-media-template/pull/136

PR #139, opened 2023-08-29
[AB#40804] make logical replication slot name empty in .env.template

AxTrusov:
[[AB#40804](https://dev.azure.com/axinom/5e61ad7f-f562-45f3-8078-9ade7df639c1/_workitems/edit/40804)]

# Pull Request

## Prerequisites

- [x] All tests pass

## Description

Because logical replication slot name has a unique nature (created on Postgres server level, is tied to a specific database, is unique, and must be defined in context of a database migration in a specific way), there is a need to make it setup a conscious decision. 

To achieve this, the following changes has been made:
- There is no default value in the `.env.template`
- For development, the `yarn db:reset` would handle filling the .env with a valid value.
- `.env.template` has an extended description of the variable

Also created a reusable function to update `.env` file with specific variables/values and reused it in all places where similar functions were defined and used.

Reminder: when the database is deleted, the replication slot is deleted as well by cascade.


services/media/service/scripts/set-replication-slot-name.ts
```diff
@@ -0,0 +1,21 @@
+/* eslint-disable no-console */
+import hash from 'short-hash';
+import { updateEnvFile } from '../../../../scripts/helpers';
+
+async function main(): Promise<void> {
+  const slotName = `media_service_localization_slot_${hash(
+    new Date().toISOString(),
```
AxTrusov, line 7: I'm using current date/time as a unique value for the hash for the following reasons:
- I could use tenantId_environmentId, but then we need to re-adjust the order of steps, filling in all the values in root .env file first, and then running yarn db:reset. Currently the order is first we do apply-templates, then db:reset, then fill in custom .env values (e.g. tenant_id, environment_id)
- I could have used the developers machine name, but then if one would have multiple instances of media-service on one machine - conflicts could happen.

In the end, current datetime seemed like an appropriate choice because:
- this value is generated only for development
- the timestamp is hashed anyway (short-hash is already being used to generate test database names, so its usability is proved by time)
- the previous slot is deleted automatically on database reset

Zehelein: > the previous slot is deleted automatically on database reset

This part is the important one to not run into too many slots issues.

Zehelein (approved):

PR #141, opened 2023-09-04
[AB#40026] Localizations integrated for TV Shows, Seasons, Episodes, Collections, and TV Show Genres

AxTrusov:
# Pull Request

## Prerequisites

- [x] All tests pass

## Description

- Localization is implemented for TV Shows, Season, Episodes, TV Show Genres, and Collections, including workflows and publishing for both media and catalog service side.
- tsconfig adjusted for backend services to change the watch mode to pre pre-4.9 state to avoid random service restars in watch mode during development on windows machines.
- Separate script `yarn util:generate-replication-slot-name` added to the media service to support with generation of a unique replication slot name for hosted scenarios.
- existing migration script `one-time-register-localization-sources.ts` adjusted to work with all localizable entities. (migration of ~8000 entities takes about 15 minutes)
- Note for the future: Workflows might require adjustments when this PR is completed: https://github.com/Axinom/mosaic-media-template/pull/124

Testing notest:

- Localization is implemented for TV Shows, Season, Episodes, TV Show Genres, and Collections in the simmilar way they are implemented for Movies.
	- Same testing shall be done - add, update localizable properties, delete, assign/unassign/change cover - these actions should be reflected in Localizations area.
	- The display title of Episodes and Seasons is generated with the same logic as for ingest item titles, shall be tested as well (episode without season; episode with season, but no tvshow; episode with season and tvshow; season without tvshow; season with tvshow)
- Known limitation: If a season exist with multiple assigned episodes, but no parent tvshow, and then this season is assigned to some tvshow - all related episodes will not have entity title updated with the localization service e.g. from `Episode 1: Test Title (Season 3)` to `Episode 1: Test Title (Season 3, TV show title)`).
   - same for when season is unassigned from TVshow or when TVshow or Season are deleted. (cascade cases would not adjust entity title of child relations)
   - this is acceptable as entity_title is a display property that is not a part of published metadata and implementing a cascade propagation is too costly.


services/media/service/scripts/generate-replication-slot-name.ts
```diff
@@ -0,0 +1,66 @@
...
+  }
+  const rl = readline.createInterface({
+    input: process.stdin,
+    output: process.stdout,
+  });
+  const def = !isNullOrWhitespace(environmentId)
+    ? ` (Default: ${environmentId})`
+    : '';
+  rl.question(
+    `> Please enter your environment ID to generate a replication slot name that is unique to it and '${serviceId}'${def}:\n`,
```
Zehelein, line 36: At this point an "ID" is not anymore required or? Should we just ask for a unique value? I mean if we ask for the service ID to be in the config we could otherwise also just ask for the env ID to be there as well - right?

services/media/service/src/domains/collections/localization/collections-images-replication-handlers.ts
```diff
@@ -0,0 +1,101 @@
...
+  return !data;
+};
+
+export const collectionsImagesReplicationHandlers = (
+  config: Config,
+  ownerPool: OwnerPgPool,
+): ReplicationOperationHandlers => {
+  const entityType = LOCALIZATION_COLLECTION_TYPE;
+  return {
+    insertHandler: async (newData: Dict<unknown> | undefined) => {
```
Zehelein, line 48: super minor: as the function is identical you might extract it as "upsertHandler" and then reuse for insert/update.

services/media/service/src/domains/tvshows/handlers/publishing-tvshow-genres-processor.ts
```diff
@@ -1,40 +1,62 @@
...
   }).run(queryable);
-
-  const snapshotJson: TvshowGenresPublishedEvent = {
-    genres: genres.map((genre) => ({
-      title: genre.title,
+  const validations: SnapshotValidationResult[] = [];
+  const mappedGenres: TvshowGenre[] = [];
+  for (const genre of genres) {
+    const { result: localizations, validation: localizationsValidation } =
+      await getTvshowGenreLocalizationsMetadata(
```
Zehelein, line 31: Maybe that could be done in a single big query to get all localizations for all TV show genres in one go?

services/media/service/src/one-time-register-localization-sources.ts
```diff
@@ -0,0 +1,437 @@
...
+  ];
+
+  const broker = await setupMessagingBroker({
+    config,
+    builders,
+    logger,
+    shutdownActions,
+  });
+
+  const ownerPool = createOwnerPgPool(
```
Zehelein, line 88: Does it actually need the owner pool for just reading data?

services/media/service/src/one-time-register-localization-sources.ts
```diff
@@ -0,0 +1,437 @@
...
+        entityId = undefined;
+        break;
+      }
+      default: {
+        continueMigration = false;
+        break;
+      }
+    }
+
+    if (typeIndex < 7) {
```
Zehelein: maybe rather `typeMapping.length`

services/media/service/src/one-time-register-localization-sources.ts
```diff
@@ -0,0 +1,437 @@
...
+          routingKey: settings.getEnvironmentRoutingKey({
+            tenantId: config.tenantId,
+            environmentId: config.environmentId,
+          }),
+        },
+      );
+      // Waiting a little for each message to be sent to make sure that all are
+      // sent before the shutdown actions are initiated, since it might take a
+      // bit of time for broker to actually send out the message.
+      await sleep(100);
```
Zehelein, line 227: Maybe this can help: https://github.com/onebeyond/rascal#flow-control to pause on "busy" and continue on "ready"? Then the sleep for a correct shutdown would not be needed.

services/media/service/scripts/generate-replication-slot-name.ts
AxTrusov: The service ID is something that is located in the service .env file, applied using an early setup step `apply-templates`, but has almost no chance of being changed/being different for different deployment environments.

Environment ID, however, will most probably be different for different environments, and will give us the uniqueness we need, and is located in root .env which is populated at later setup step.

The main target of this script is to get a value for hosting scenarios (to go into the hosting yaml file). You are of course not limited to entering only env ids, you can enter anything and would still get a usable value, but with env ID + service ID we are at least giving a clear indication of how the resulting value uniqueness is ensured.

This script is something we agreed with Ruwan and he seemed fine with it, so I think the current version is ok for now.

services/media/service/src/domains/tvshows/handlers/publishing-tvshow-genres-processor.ts
AxTrusov: This was already brought up in the Movies PR. This would require some investigation/custom processing/generation of queries. There is a separate task for this: #40874

services/media/service/src/domains/collections/localization/collections-images-replication-handlers.ts
AxTrusov: We already have `getUpsertMessageData`, so I would say creating an `upsertHandler` would not improve it too much and might actually lose some readability. I think I'd like to keep it as is for now, it would also make it easier to add possible customizations in the future.

services/media/service/src/one-time-register-localization-sources.ts
AxTrusov: we can use login pool + gql role but then we would need to establish transactions explicitly for each operation (currently the zapatos is doing it). Since this is a migration script - using owner pool is acceptable.

services/media/service/src/one-time-register-localization-sources.ts
AxTrusov: the current broker instance is a wrapper around the `rascal` broker and it does not expose the `on` handler. Discussed internally, the current implementation is acceptable for a one-time script. For a proper implementation - a more sophisticated way to detect if all messages have been sent would have been required.

Zehelein (approved):

PR #155, opened 2023-09-16
[AB#41434] Add a quick-deploy interactive script to media-template

EmperorRXF:
- We can now run `yarn quick-deploy` from the package root to start the interactive deployment script
- The script will perform some basic validations (i.e. if docker CLI is installed, the user is logged in, the deployment manifests have been updated with user specific variables, etc.)

- Screencap of usage:
https://github.com/Axinom/mosaic-media-template/assets/8752394/6fcf1a04-a126-4e41-ac1a-2f9a94b37402

- Testing Notes:
  - Follow the "Quick Deploy" section updated in the README file of the package root


scripts/quick-deploy-service.ts
```diff
@@ -0,0 +1,393 @@
...
+        type: 'text',
+        name: 'dockerImageTag',
+        message:
+          'If you want to deploy a specific docker image tag that is already pushed to the registry, enter that image tag here. Leave this blank if you you are not sure.',
+        initial: '',
+      },
+      {
+        type: 'confirm',
+        name: 'deploy',
+        message: 'Confirm the deployment of the selected service?',
```
BenjaminSchwendner: this was a bit confusing in the video tbh
wasn't sure what it was configuring and was guessing with this to `yes` it would just build and then ask again before it start the deployment. But from code, I see now, that this is essentially an "are you sure you want to start deploying?" prompt. Which means, we should probably also phrase it like this?

EmperorRXF: brain is a funny thing at interpreting short instructions ;) fixed now :)

scripts/quick-deploy-service.ts
```diff
@@ -0,0 +1,393 @@
...
+
+/**
+ * Validate if the deployment manifest YAML file of the service still has some variables that need to be manually adjusted.
+ */
+function validateDeploymentManifestIsModified(
+  serviceId: 'media-service' | 'catalog-service' | 'entitlement-service',
+): void {
+  const servicePrefix = serviceId.split('-service')[0];
+
+  if (serviceId === 'entitlement-service') {
```
yasithA: Any reason why we're only doing this for `entitlement-service` but accept arguments for all other services as well? Feel like the validation below, of checking for `REPLACE_WITH_VALID_` is generic enough for all services?

yasithA (approved):

scripts/quick-deploy-service.ts
EmperorRXF: no actually - this was probably a remnant filter condition I had while evolving the script. it should be generic enough to work for all services given they all follow the same naming convention for client replicable variables.

thanks!

PR #158, opened 2023-09-27
[AB#36747] Use ID service availability check on customizable services

axinom-nrathnayaka:
# Pull Request

## Prerequisites

- [x] All tests pass

## Description

For Release Notes:

- A new method setupServiceHealthEndpoint is introduced to the @axinom/mosaic-service-common library. It exposes a /health route on the service API URL, that can be used to check for a HTTP 200 OK response to determine if a service is available for use. Optionally, the response body may also contain an arbitrary JSON document specified by the service that describes the health of individual components of the service. See https://github.com/Axinom/mosaic-media-template/pull/158 for more details on its usage.
- A new method isServiceAvailable is introduced to the @axinom/mosaic-service-common library. This method will query the route configured via the setupServiceHealthEndpoint method to check if the target service is available for use. If the response contains a HTTP 200 response code, it will return true, else will return false. The method will retry the availability check for a default 15 attempts with an exponential back-off strategy before returning false. See https://github.com/Axinom/mosaic-media-template/pull/158 for more details on its usage.

### DIFF
--- a/services/catalog/service/src/index.ts
+++ b/services/catalog/service/src/index.ts
@@ -14,12 +14,14 @@ import {
 import {
   closeHttpServer,
   handleGlobalErrors,
+  isServiceAvailable,
   Logger,
   MosaicErrors,
   setupGlobalConsoleOverride,
   setupGlobalLogMiddleware,
   setupLivenessAndReadiness,
   setupMonitoring,
+  setupServiceHealthEndpoint,
   setupShutdownActions,
   tenantEnvironmentIdsLogMiddleware,
 } from '@axinom/mosaic-service-common';
@@ -40,6 +42,16 @@ async function bootstrap(): Promise<void> {
   const app = express();
 
   const { readiness } = setupLivenessAndReadiness(config);
+
+  // Check ID service is available
+  if (!(await isServiceAvailable(config.idServiceAuthBaseUrl))) {
+    throw new Error(
+      'The Mosaic Identity Service is required to run this service, but it was not available.',
+    );
+  }
+
+  // Register service health endpoint
+  setupServiceHealthEndpoint(app, config.port);
 
   await applyMigrations(config);
   const shutdownActions = setupShutdownActions(app, logger);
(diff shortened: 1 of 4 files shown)


services/catalog/service/src/index.ts
```diff
@@ -41,6 +43,16 @@ async function bootstrap(): Promise<void> {
 
   const { readiness } = setupLivenessAndReadiness(config);
 
+  // Check ID service is available
+  if (!(await isServiceAvailable(config.idServiceAuthBaseUrl))) {
```
EmperorRXF, line 47: catalog-service does not need id-service to run. so no need for any check here

EmperorRXF: ok we use RMQ messages in catalog-service, and id-service is needed to validate JWTs. we can keep this check for that

EmperorRXF:
pls add this to PR description:

For Release Notes:
- A new method `setupServiceHealthEndpoint` is introduced to the `@axinom/mosaic-service-common` library. It exposes a `/health` route on the service API URL, that can be used to check for a HTTP 200 OK response to determine if a service is available for use. Optionally, the response body may also contain an arbitrary JSON document specified by the service that describes the health of individual components of the service. See [this PR](https://github.com/Axinom/mosaic-media-template/pull/158) for more details on its usage.
- A new method `isServiceAvailable` is introduced to the `@axinom/mosaic-service-common` library. This method will query the route configured via the `setupServiceHealthEndpoint` method to check if the target service is available for use. If the response contains a HTTP 200 response code, it will return `true`, else will return `false`. The method will retry the availability check for a default 15 attempts with an exponential back-off strategy before returning `false`. See [this PR](https://github.com/Axinom/mosaic-media-template/pull/158) for more details on its usage.

EmperorRXF (approved):

EmperorRXF:
PR should be raised for dev branch

axinom-nrathnayaka:
Will abandon this PR, another PR had raised instead of this.

PR #164, opened 2023-10-05
[AB#41699] Localizations integration

AxTrusov:
# Pull Request

## Prerequisites

- [x] All tests pass

## Release Notes

This PR introduces localization support for all publishable entities (Movie, Tv Show, Season, Episode, Collection, Movie Genre, Tv Show Genre).

- This PR changes the Catalog service by moving localizable properties into a separate related table, e.g. title, description, and synopsis are no longer part of the `movie`, but are moved to `movie_localizations` related table. 
   - The GraphQL API remains unchanged, as dedicated views are implemented to expose localizable properties on the level of parent entity.
- Localization is an optional feature in the media service and can be enabled/disabled using `LOCALIZATION_IS_ENABLED` environment variable.
    - When disabled, entity definitions and changes from the media service will not be synced with the localization service, and only base values will be published instead of base values and localizations.
- After these changes are applied to an existing media-template repository and there is an intention to enable localizations, the following migration steps must be done:
   - Existing media-service service account must be adjusted to include `Source Entities: Edit`, `Entity Definitions: Edit`, `Localized Entities: Edit`, and `Localized Entities: Review` permissions of the localization service, and `Ingests: Edit` for the media service itself. For local development, this can be done by running `yarn setup` script to (re)create the service account.
   - Once the Localization service is enabled in both Admin Portal and Media Service (via `LOCALIZATION_IS_ENABLED` env var) and successfully launched, previously existing entities can be migrated to become localizable using a dedicated migration script located at `services\media\service\src\_manual-repair-scripts\register-localization-sources.ts`

More details in:
- [Localization Service](https://portal.axinom.com/mosaic/documentation/media/localization-service)
- [Ingest of Localizations PR](https://github.com/Axinom/mosaic-media-template/pull/211)


Zehelein (approved):
Thanks

AxTrusov:
adding a soft-blocker to merge this:

since we encountered some issues with DB size growth with WAL, we should wait until we discuss in the PA round to understand risks/potential workarounds before releasing this change to the public template

PR #165, opened 2023-10-06
[AB#36747] Use ID service availability check on customizable services

axinom-nrathnayaka:
# Pull Request

## Prerequisites

- [x] All tests pass

## Description
- `setupServiceHealthEndpoint` is used to register a health endpoint in each customizable-service.
- `isServiceAvailable` is used to ensure Mosaic ID Service is available on service bootstrap.

## For Release Notes:
- A new method `setupServiceHealthEndpoint` is introduced to the `@axinom/mosaic-service-common` library. It exposes a `/health` route on the service API URL, that can be used to check for a HTTP 200 OK response to determine if a service is available for use. Optionally, the response body may also contain an arbitrary JSON document specified by the service that describes the health of individual components of the service. See [this PR](https://github.com/Axinom/mosaic-media-template/pull/158) for more details on its usage.
- A new method `isServiceAvailable` is introduced to the `@axinom/mosaic-service-common` library. This method will query the route configured via the `setupServiceHealthEndpoint` method to check if the target service is available for use. If the response contains a HTTP 200 response code, it will return `true`, else will return `false`. The method will retry the availability check for a default 15 attempts with an exponential back-off strategy before returning `false`. See [this PR](https://github.com/Axinom/mosaic-media-template/pull/158) for more details on its usage.

## For Testing Notes:
- You can send request to /health end point and can check that endpoint is working. Can check for any manage service, not only for id-service. 
- This is for id-service,
Ex: https://id.service.test.axinom.net/health
- Should return empty json object with 200 ok response code.

### DIFF
--- a/services/catalog/service/src/index.ts
+++ b/services/catalog/service/src/index.ts
@@ -14,12 +14,14 @@ import {
 import {
   closeHttpServer,
   handleGlobalErrors,
+  isServiceAvailable,
   Logger,
   MosaicErrors,
   setupGlobalConsoleOverride,
   setupGlobalLogMiddleware,
   setupLivenessAndReadiness,
   setupMonitoring,
+  setupServiceHealthEndpoint,
   setupShutdownActions,
   tenantEnvironmentIdsLogMiddleware,
 } from '@axinom/mosaic-service-common';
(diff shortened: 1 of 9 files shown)


services/catalog/service/src/index.ts
```diff
@@ -41,6 +43,16 @@ async function bootstrap(): Promise<void> {
...
 
+  // Check ID service is available
+  if (!(await isServiceAvailable(config.idServiceAuthBaseUrl))) {
+    throw new Error(
+      'The Mosaic Identity Service is required to run this service, but it was not available.',
+    );
+  }
+
+  // Register service health endpoint
+  setupServiceHealthEndpoint(app, config.port);
```
EmperorRXF: think we removed the 2nd arg of the `setupServiceHealthEndpoint`, and only require the `app` object now?

pls ensure mosaic-service-common is the latest RC version that has this new method signature

EmperorRXF (approved):

PR #166, opened 2023-10-09
[AB#41638] Add docker username as arg to quick deploy script

yasithA:
# Pull Request

## Prerequisites

- [x] All tests pass

## Description

Add docker username as arg to quick deploy script

### DIFF
--- a/scripts/quick-deploy-service.ts
+++ b/scripts/quick-deploy-service.ts
@@ -180,7 +180,7 @@ Please follow the service README file to understand how to generate the values, 
   }
 }
 
-function getDockerInfo(): { registry: string; username?: string } {
+function getDockerInfo(): { registry?: string; username?: string } {
   console.log(`\nChecking Docker Info...\n`);
 
   try {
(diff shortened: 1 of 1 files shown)


scripts/quick-deploy-service.ts
```diff
@@ -347,24 +363,29 @@ async function main(): Promise<void> {
     } else {
       const uniqueID = getUniqueID();
 
-      const { registry, username } = getDockerInfo();
-
-      console.log(`Docker Registry: ${chalk.green(registry)}`);
-      console.log(`Docker Username: ${chalk.green(username)}`);
+      console.log(`Docker Username: ${chalk.green(answers.dockerUsername)}`);
```
EmperorRXF: this we can remove this?

scripts/quick-deploy-service.ts
```diff
@@ -326,6 +336,12 @@ async function main(): Promise<void> {
           { title: 'Entitlement Service', value: 'entitlement-service' },
         ],
       },
+      {
+        type: 'text',
+        name: 'dockerUsername',
+        message: 'Enter the Docker username',
+        initial: username,
```
EmperorRXF, line 343: if the username is not supplied, can we handle that scenario bit better?

either make `prompt` not proceed with "empty" values or we fail early with an error - rather than proceeding upto build step and then only giving some error :)

EmperorRXF (approved):

PR #181, opened 2023-10-18
[AB#41785] separate replication connection string for deployment scenarios

AxTrusov:
# Pull Request

## Prerequisites

- [x] All tests pass

## Description

When service is deployed, the logical replication service must connect to the database directly and not through pg-bouncer. For this reason, a separate port value and owner connection string are used to make sure only replication services are connecting to the database avoiding the pg-bouncer.

## Testing Notes

Make sure that localizations work. (e.g. creating a movie will also create a movie entry in localizations workflow)


.env.template
```diff
@@ -35,6 +35,7 @@ ID_SERVICE_LOCAL_PROXY_PORT=10505
 # Database Connection
 POSTGRESQL_HOST=localhost
 POSTGRESQL_PORT=5432
+POSTGRESQL_REPLICATION_PORT=5432
```
Zehelein, line 42: Here we would need to describe that the above is/can use PG bouncer and the replication one could bypass pg bouncer or similar tool and directly connect to the DB. The DB/DBMS is the same in both cases.

Zehelein (approved):

PR #187, opened 2023-10-26
[AB#41880] localizable properties are embedded into the root entity

AxTrusov:
# Pull Request

## Prerequisites

- [x] All tests pass

## Description

Instead of having a `localization` sub-object for each localizable entity, localizable properties are now embedded into it.
The filtering by locale is happening via the headers not to break "[the contract that a given identified entity is always the same entity](https://discord.com/channels/489127045289476126/498852330754801666/1166690921547571240)". 

This implementation makes it more convenient for the FE apps that would be using the Catalog Service API, since there will be no need for them to know which property is localizable, and which is not. But this comes at the cost of a performance drop (due to N+1 issue).

## Testing notes
Try testing the catalog API and make sure that content is returned based on the passed header parameter, e.g.
```
{
"mosaic-locale": "de-DE"
}
```
The general behavior should remain the same (e.g. if locale exists - return the localized value. If it does not exist or not specified - base value is returned). Localizable properties include title, description, synopsis. Channel has no synopsis, Season has no title, Genres have only title.


services/catalog/service/src/graphql/plugins/entity-localization-plugin.ts
```diff
@@ -0,0 +1,83 @@
...
+  embed,
+  gql,
+  makeExtendSchemaPlugin,
+  makePluginByCombiningPlugins,
+} from 'graphile-utils';
+
+import { Table } from 'zapatos/schema';
+import { MOSAIC_LOCALE_PG_KEY } from '../../common';
+
+const wellKnownProperties = [
```
Zehelein: Instead of an array you may use an object then you do not need "find" to search through the array but could access it via `wellKnownProperties[property]`

services/catalog/service/src/graphql/plugins/entity-localization-plugin.ts
```diff
@@ -0,0 +1,83 @@
...
+      ? `"""${fieldDefinition.description(entityName)}"""`
+      : '';
+    const required = fieldDefinition?.required ? `!` : '';
+    const capitalizedEntityName = inflection.upperCamelCase(entityName); // e.g. Movie
+    const localizations_table = sql.identifier(localizationTable);
+    const schema = sql.identifier('app_public');
+    const column = sql.identifier(property);
+    const fk = sql.identifier(`${entityName}_id`);
+    const localeKey = sql.value(MOSAIC_LOCALE_PG_KEY);
+    const condition = sql.fragment`(${schema}.${localizations_table}.locale = (SELECT pg_catalog.current_setting(${localeKey}, true)) OR ${schema}.${localizations_table}.is_default_locale IS TRUE)`;
```
Zehelein: minor: why extract this condition and not include it or extract other conditions as well?

services/catalog/service/src/tests/tvshow-genres/localizations.db.spec.ts
```diff
@@ -71,18 +77,18 @@ describe('Tvshow genre Localization Graphql Requests', () => {
     }).run(ctx.ownerPool);
 
     // Act
-    const resp = await ctx.runGqlQuery(TVSHOW_GENRE_REQUEST, {
-      locale: '',
-    });
+    const resp = await ctx.runGqlQuery(
+      TVSHOW_GENRE_REQUEST,
+      {},
+      getRequestContext('asdf'),
```
Zehelein: filter should be empty here

Zehelein (approved):

services/catalog/service/src/common/constants.ts
```diff
@@ -1 +1,9 @@
 export const DEFAULT_LOCALE_TAG = 'default';
+/**
+ * The HTTP Request header key for locale filtering.
+ */
+export const MOSAIC_LOCALE_HEADER_KEY = 'mosaic-locale';
```
BenjaminSchwendner, line 5: shouldn't this be something like `X-`?

did we think about using the `Accept-Language` header?

Zehelein: "Custom proprietary headers have historically been used with an X- prefix, but this convention was deprecated in June 2012" (https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers)

Zehelein: I was listening to the GraphQL specification steering videos and there is the GraphQL over HTTP specification in the works: https://github.com/graphql/graphql-over-http/blob/main/spec/GraphQLOverHTTP.md They said this captures the "what is already done in the wild" and offers an "extensions" object. Not sure if this is better (and we should also send the JWT in there) - I think there is also no GraphQL strong type support for that object.

BenjaminSchwendner: ok, not an `X-` then :-)

About GraqphQL over HTTP - we can do that, once it's getting somewhere.

Question for now would be, should we use the `Accept-Language` header? Maybe we could use our header as preferred, and if it's not present, we try to find a locale for the `Accept-Language`?

AxTrusov: For simplicity's sake, I would stick with only a custom header. Since this is a customizable service - this can be adjusted based on the needs of the custom frontends.

AxTrusov:
Closed in favor of an approach with views. See #193

PR #192, opened 2023-11-07
[AB#41558] max pool connections increased from 10 to 25

AxTrusov:
# Pull Request

## Prerequisites

      releases)

## Description

- `PG_POOL_MAX_CONNECTIONS` config variable was added to backend services to increase the default limit of 10 connections to 25.
- Most services adjusted to pass login pool as a parameter to `postgraphile(` middleware instead of the connection string. 
   - this made sure that we do not have 2 login pools in circulation and increased connections limit is applied to regular postraphile-generated operations

## Testing notes

Smoke tests should be enough.


services/catalog/service/mosaic-hosting-deployment-manifest.yaml
```diff
@@ -30,6 +30,7 @@ regularVariables:
   DATABASE_LOGIN: '${__ax_hosted__.pg.db_login_role}'
   DATABASE_LOGIN_PASSWORD: '${__ax_hosted__.pg.db_login_password}'
   DATABASE_GQL_ROLE: '${__ax_hosted__.pg.db_gql_role}'
+  PG_POOL_MAX_CONNECTIONS: '${__ax_hosted__.pg.pool_max_connections}'
```
yasithA: Had a chat with @EmperorRXF as well, and we can remove the placeholder here and get the media-template developers to straightaway mention a value, since it's something they can decide and the Hosting Service does not have to interfere with it. I shall push the fix shortly. :)

yasithA (approved):

PR #194, opened 2023-11-09
[AB#41993]Added aggregate type and ID to messages

Zehelein:
# Pull Request

## Prerequisites

      releases)
- [X] potential **release notes** to the PR description added

## Release Notes

- Regenerated the MessagingSettings based on the latest Moscai library updates
- Updated the broker messaging publication code to include the aggregate ID and aggregate type.

## Testing Notes

- Make sure all messaging-related code still works: ingest, publishing and unpublishing genres, movies, TV shows, seasons, episodes, collections, and channels


libs/media-messages/schemas/payloads/media/media-service-asyncapi.yml
```diff
@@ -86,38 +86,56 @@ channels:
...
         $ref: 'commands/delete-entity-command.json'
     publish-entity-command:
+      tags:
+        - name: aggregate-type:entity
       contentType: application/json
       payload:
         $ref: 'commands/publish-entity-command.json'
     start-ingest-command:
+      tags:
+        - name: aggregate-type:ingest
```
AxTrusov: I think it would be more accurate to change to `aggregate-type:ingest-document`
in both cases where this is used (to align with `ingest-item`)

libs/media-messages/schemas/payloads/media/media-service-asyncapi.yml
```diff
@@ -86,38 +86,56 @@ channels:
...
         $ref: 'commands/start-ingest-item-command.json'
     unpublish-entity-command:
+      tags:
+        - name: aggregate-type:entity
       contentType: application/json
       payload:
         $ref: 'commands/unpublish-entity-command.json'
     update-metadata-command:
+      tags:
+        - name: aggregate-type:ingest-item
```
AxTrusov, line 132: Very minor, but can also be `ingest-item-step` (we do process each step separately)

AxTrusov: Actually, to align with
`services\media\service\src\ingest\handlers\start-ingest-item-handler.ts` (using
entity_id as an aggregate-id) - this should be `aggregate-type:entity`.

AxTrusov (approved):

libs/media-messages/schemas/payloads/media/media-service-asyncapi.yml
Zehelein: Good point - adjusted the code to use the item ID instead to rather be in line with the other ingests.

PR #199, opened 2023-11-17
[AB#42068] Add a filter option for ingest results

BenjaminSchwendner:
# Pull Request

## Prerequisites

      releases)
- [x] potential **release notes** to the PR description added

## Description

# Release Notes
Introducing a filter on the ingest details station that will allow filtering the list by status, in case there are multiple statuses present.


axisuru:
There's a 5000ms poll interval which fetches new data for the Station. This interval also resets the filters. 

https://github.com/Axinom/mosaic-media-template/assets/92033440/798572cd-c334-4952-a7b6-e3f779f2e082

BenjaminSchwendner:
> There's a 5000ms poll interval which fetches new data for the Station. This interval also resets the filters.
> 
>  Screen.Recording.2023-11-17.at.23.01.06.mov

Pushed an update that should fix this issue. 
Though, for whatever reason, it did not behave like this on my machine. But this is likely something wrong with my setup, because the polling did just not trigger a re-render. Which should of course be the case.

axisuru (approved):

PR #211, opened 2023-12-05
[AB#40605] Ingest of localizations is now supported

AxTrusov:
# Pull Request

## Prerequisites

- [ ] The PR is targeting the right branch (`dev` for features and `master` for
      releases)

## Description

Ingest of localizations is now supported for movies, tvshows, seasons, and episodes.
- Ingest json format adjusted to include an optional `localizations` array that can look like this.
```
localizations: [
  {
    "language_tag": "de-DE",
    "title": "Avatar – Aufbruch nach Pandora",
    "description": "Avatar ist ein US-amerikanischer Science-Fiction-Epos aus dem Jahr 2009...",
    "synopsis": "Im Jahr 2154 haben die Menschen die natürlichen Ressourcen der Erde erschöpft..."
  },
  {
    "language_tag": "et-EE",
    "title": "Avatar",
    "description": "Avatar on 2009. aasta Ameerika eepiline ulmefilm...",
    "synopsis": "Aastal 2154 on inimesed ammendanud Maa loodusvarad..."
  }
]
```
- This array should only contain objects for non-default languages that are configured in the Admin Portal for the Localization service.
   - If present - other languages will still be sent - but will be ignored by the Localization service.
- The synchronization of localization source entities happens using logical replication handlers and is separate from the ingest process.
	- This means that special handling is done to make sure that first, source entity is synchronized with the localization service, and only after that the localizations are sent out to the localization service.
	- To achieve this sequence of events and to avoid concurrency issues, the following was done:
	   - Localizable and ingestable entities now have a new column `ingest_correlation_id` (e.g. in `movies` table)
	   - When ingest handles the `UpdateMetadata` step - this new column value is set to the ingest item id, which triggers the logical replication handler with an ingest context.
	      - This is done only if localizations are enabled for the service and actual localizations are provided for the item (e.g. movie) in the ingest document.
	   - The point above makes sure that Logical replication handlers would include the ingest-specific context in the sent `UpsertLocalizationSourceEntity` command message.
		  - `ingest_correlation_id` is set to null in a separate update in context of the same transaction. This ensures that only one specific update will include extra ingest context. 
		  - This also means, that this is purely a system column that in database will always have NULL as a value and it is not exposed in the GraphQL API.
	   - Usually, `UpsertLocalizationSourceEntity` would not produce events to avoid unnecessary spam, but when any context is provided - it will send out success and failure events (with previously provided context)
	   - Media service will receive these events (with `ingestItemId` in the `message_context`) and use this as an indication that localization service is now ready to receive the actual localizations for that specific ingest item (e.g. movie), and proceed to send such message.
- ingest-gen script is adjusted to support localizations
- setup script is adjusted to include new permissions for service account
	- hosting manifest includes new localization-service permissions, but media-service permissions must be set manually. This is temporary workaround and would be improved in the future.
- The overall sequence of messages that lead to the completion of the newly added `LOCALIZATIONS` ingest item step (other steps are omitted to simplify this):
   - Ingest document is uploaded -> `StartIngestCommand` ->  `StartIngestItemCommand` -> `UpdateMetadata` -> `UpsertLocalizationSourceEntityCommand` (logical replication) -> `UpsertLocalizationSourceEntityFinishedEvent` -> `LocalizeEntityCommand` -> `LocalizeEntityFinishedEvent` -> `CheckFinishIngestItemCommand`

## Testing notes
- (one-time check) Make sure that Localizations are enabled for the media service, that Localization service is enabled in the Admin Portal, and Localization service has configured non-default locales in the Admin Portal.
- (one-time check) Check the Media Service service account in the Admin Portal. It should have `SOURCE_ENTITIES_EDIT`, `ENTITY_DEFINITIONS_EDIT`, `LOCALIZED_ENTITIES_REVIEW` permissions for the Localization service and `INGESTS_EDIT` permission for the media service itself.
- Create or adjust the ingest document json so that ingest items would include `localizations` array similar to the example above. 
   - `ingest-gen` script in the media service can help with generating a document with random data, check media-service README.md for details
- Start ingest with prepared ingest document json
- observe the status of ingest items in the UI 
   - Each item that has a `localizations` array should have a `LOCALIZATIONS` step (with empty subtype, e.g. IMAGE could have `COVER` sub-type, but `LOCALIZATIONS` step would not have one)
   - At some point in time - the `LOCALIZATIONS` step should have a `SUCCESS` state.
- Observe the localizations workflows (either from home or from details of ingested entities) - localizations should be set for provided languages and fields with status `APPROVED`.


services/media/service/src/ingest/handlers/localize-entity-failed-handler.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+        ingestItemStepId: '34d91ea5-db63-4e51-b511-ae545d5c669c',
+        ingestItemId: 1,
+        imageType: 'MAIN',
+      });
+
+      // Act
+      await handler.onMessage(content, message);
+
+      // Assert
+      expect(messages[0]).toEqual<CheckFinishIngestItemCommand>({
```
Zehelein, line 67: tiny addition: maybe check that there is exactly one message?

services/media/service/src/ingest/handlers/localize-entity-failed-handler.ts
```diff
@@ -0,0 +1,60 @@
...
+      ],
+    );
+  }
+
+  async onMessage(
+    content: LocalizeEntityFailedEvent,
+    message: MessageInfo,
+  ): Promise<void> {
+    if (content.service_id !== this.config.serviceId) {
+      // skipping message from different service
```
Zehelein: Maybe include that it is about the entity types?

> skipping events for entity types from different services

services/media/service/src/ingest/handlers/upsert-localization-source-entity-finished-handler.ts
```diff
@@ -0,0 +1,182 @@
...
+      localizations,
+    };
+
+    const localizationMessageContext: IngestMessageContext = {
+      ingestItemId: messageContext.ingestItemId,
+      ingestItemStepId: localizationStep.id,
+    };
+
+    await this.broker.publish<LocalizeEntityCommand>(
+      UNKNOWN_AGGREGATE_ID,
```
Zehelein: Maybe set the entity ID here as  `content.entity_id`? The aggregate type is "localization" so obviously it would not be the only key. But it would IMO still be the best thing to set there because it is not really unknown but rather just "not enough to exactly specify the entity without also using values from the payload".

services/media/service/src/ingest/schemas/ingest-validation-schema.json
```diff
@@ -175,6 +175,21 @@
           }
         }
       }
+    },
+    "localizations": {
+      "type": ["array", "null"],
+      "description": "Localizations for every defined locale.",
```
Zehelein: Likely rather "Localizations for any defined locale." or? You don't need to provided all of them (at least it is also possible to send in "null" instead of an array).

Zehelein (approved):

PR #219, opened 2023-12-19
[AB#42832]Implemented the transactional outbox and inbox pattern for the media service

Zehelein:
# Pull Request

## Prerequisites

      releases)
- [X] potential **testing notes** to the PR description added

## Description

This PR improves the messaging reliability by applying the transactional outbox and inbox patterns (e.g. https://softwaremill.com/microservices-101/). Messages are not sent immediately but first stored in the database together with the surrounding database transaction. So both message handling and message sending succeeds or none at all. Similar for the inbox side: either all the processing succeeds as well as marking the inbox message as handled - or nothing succeeds. 

## Testing Notes

For testing, you can on the one hand test that messages continue to work. Examples are to deploy a new media service into a new environment and check that the image types and cue point types are registered in the image/video service. And that bulk archive and unarchive as well as publishing entities to the catalog service works.

You can check the logs that no transactional-related errors are logged (only the outbox/inbox ones and message handled etc.).

In the database, you can check the "app_hidden.inbox" and "app_hidden.outbox" tables and ensure that there are no entries that have the `abandoned_at` date set and in general that the `started_attempts` numbers are most of the time "1". A few might be retried and have higher numbers but if those are only few this is fine - it is basically why we implement the transactional outbox and inbox logic.


libs/media-messages/schemas/payloads/media/media-service-asyncapi.yml
```diff
@@ -79,7 +79,7 @@ channels:
     bindings:
       amqp:
         queue:
-          name: entity:deleted
+          name: inbox
```
AxTrusov, line 82: Just to make sure, should this also be changed, even though this is an event and
not a command?

services/media/service/.env.template
```diff
@@ -13,6 +13,10 @@ DATABASE_LOGIN=media_service_login
 DATABASE_LOGIN_PASSWORD=media_service_login_pwd
 DATABASE_GQL_ROLE=media_service_gql_role
 
+# NOTE: If these values change the old database replication slots must be explicitly dropped.
```
AxTrusov: In the `feature/localization` branch - we explicitly left slot env var value empty in the
.env.template and incorporated a few script changes to:
- make db:reset generate explicit unique slot names
- added a new script to be able to generate unique values for deployment.

If the current state is left and 2 different partners would deploy their
versions of media-template - second one would fail, because default replication
slot names would be used.

`feature/localization` branch also has a huge comment here describing this
stuff. Perhaps something similar can be made here.

services/media/service/.env.template
```diff
@@ -13,6 +13,10 @@ DATABASE_LOGIN=media_service_login
...
+TRANSACTIONAL_OUTBOX_REPLICATION_SLOT=ott_media_service_transactional_outbox_slot
```
AxTrusov: Please also adjust the mosaic-hosting-deployment-manifest.yaml. Check
`feature/localization` branch for example, as we added slot name as empty and
advice using comment to generate a unique name using script (already mentioned
in previous comment)

services/media/service/.env.template
```diff
@@ -13,6 +13,10 @@ DATABASE_LOGIN=media_service_login
...
+TRANSACTIONAL_INBOX_REPLICATION_SLOT=ott_media_service_transactional_inbox_slot
```
AxTrusov: Because it's not clear which PR will be completed first - this, or
`feature/localization` - this PR might still require some changes that are
already in the `feature/localization`
- Root .env.template - add `POSTGRESQL_REPLICATION_PORT`
- adjust `infra/docker-compose.yml` to include replication parameters

yarn.lock
```diff
@@ -192,16 +192,16 @@
   dependencies:
     "@types/json-schema" "^7.0.11"
 
-"@axinom/mosaic-cli@0.26.0-rc.0":
```
AxTrusov: Please re-run `yarn`, `yarn.lock` is currently not in its latest state

EmperorRXF:
adding a soft-blocker to merge this:

- since we encountered some issues with DB size growth with WAL, we should wait until we discuss in the PA round to understand risks/potential workarounds before releasing this change to the public template

scripts/helpers/db-test-helpers.ts
```diff
@@ -65,6 +65,7 @@ export async function recreateTestDbTemplate(
     dbConfig.dbLoginPassword,
     dbConfig.dbOwner,
     dbConfig.dbOwnerPassword,
+    true,
```
AxTrusov: to follow the approach with db:reset adjustments, this should also be a
parameter, so that e.g. `test:reset:dbs` would not create test owner users with
replication attribute for catalog and entitlement services

services/media/service/src/messaging/register-messaging.ts
```diff
@@ -1,39 +1,348 @@
...
+): Promise<Broker> => {
+  const [storeInboxMessage, shutdownInbox] = initializeInboxMessageStorage(
+    inboxConfig,
+    logMapper,
+  );
+  shutdownActions.push(shutdownInbox);
+  const inboxWriter = new RabbitMqInboxWriter(storeInboxMessage, {
+    // temporary backward compatibility until all your services are updated and all current messages are processed
+    acceptedMessageSettings: [
+      MediaServiceMessagingSettings.StartIngest,
```
AxTrusov: Does this affect only ingest-related messages, or message types from
`publishingBuilders` and `commonBuilders` should be included as well?

services/media/service/src/ingest/handlers/check-finish-ingest-document-handler.ts
```diff
@@ -100,41 +89,39 @@ export class CheckFinishIngestDocumentHandler extends MediaGuardedMessageHandler
...
+      await update(
+        'ingest_documents',
+        {
+          status: 'ERROR',
+          errors: sql<SQL>`${value} || ${error}::jsonb`,
         },
+        { id: ingest_document_id },
+      ).run(loginClient);
+    } else {
+      await sleep(5000);
```
AxTrusov: This now bothers me... previously we did not hold the transaction open for 5
seconds, but now this code does. Is there any way to complete the transaction and
then wait? or mark the sent `CheckFinishIngestDocument` message as delayed?

services/media/service/src/ingest/handlers/image-failed-handler.ts
```diff
@@ -1,55 +1,59 @@
...
+  async handleMessage(
+    {
+      payload,
+      metadata,
+    }: TransactionalInboxMessage<EnsureImageExistsFailedEvent>,
+    loginClient: ClientBase,
   ): Promise<void> {
-    const messageContext = message.envelope
-      .message_context as ImageMessageContext;
+    const messageContext = metadata.messageContext as ImageMessageContext;
```
AxTrusov, line 49: This message should also use `checkIsIngestEvent`

services/media/service/src/ingest/handlers/start-ingest-item-handler.ts
```diff
@@ -21,101 +23,92 @@ import {
...
-          ...original,
-          publicationConfig: this.getPublicationConfig(
-            original.messagingSettings,
-          ),
-        }));
-      },
-    );
+    const data = processor.getOrchestrationData(payload);
+    await this.saveIngestItemSteps(data, loginClient);
+    const orchestrationData = await data.map((original) => ({
```
AxTrusov: await not needed here

services/media/service/src/ingest/handlers/video-failed-handler.spec.ts
```diff
@@ -44,26 +48,28 @@ describe('VideoFailedHandler', () => {
...
-      expect(messages[0]).toEqual<CheckFinishIngestItemCommand>({
-        ingest_item_step_id: '8331d916-575e-4555-99da-ac820d456a7b',
-        ingest_item_id: 1,
-        error_message: 'Test error message',
-      });
+      handler.handleMessage(
+        createMessage(payload, context),
+        stub<ClientBase>(),
+      ),
+        // Asert
```
AxTrusov: one `s` got lost in `Assert`

services/media/service/src/publishing/plugins/entity-publishing-endpoints-plugin-factory.ts
```diff
@@ -93,36 +93,17 @@ export const EntityPublishingEndpointsPluginFactory = (
...
-                  jwtToken,
-                  config,
-                  messagingBroker,
-                } = getValidatedExtendedContext(context);
-                const pgSettings = buildPgSettings(
-                  subject,
-                  config.dbGqlRole,
-                  config.serviceId,
-                );
+                const { jwtToken, config, pgClient, storeOutboxMessage } =
```
AxTrusov, line 96: While using `pgClient` makes things work totally fine atm, I would still prefer
to use explicit transactions in all postgraphile custom plugins where we perform
writes and only use `pgClient` if we need to only make selects.

Reason - postgraphile will commit `pgClient` transaction if an error is thrown
within custom plugin and after some db operation. Postgraphile even suggests to
use `Savepoints` to aleviate this.

The case above will currently not happen, because we use
`AtomicMutationsPlugin`, which silently "fixes" this. 
But if you were to remove `AtomicMutationsPlugin` from postgraphile option 
and add something like `throw new Error('test');` on the line 121 (after
`storeOutboxMessage`) and go to Movie Properties of some publishable movie and
click `Publish Now` - you would see an explicitly thrown error - but movie would
still be published.

services/media/service/src/domains/common/handlers/image-types-declared-handler.ts
```diff
@@ -0,0 +1,32 @@
...
+        context: ImageTypesDeclaredHandler.name,
+      }),
+      config,
+    );
+  }
+
+  async handleMessage({
+    payload,
+  }: TransactionalInboxMessage<ImageTypesDeclaredEvent>): Promise<void> {
+    this.logger.log({
```
AxTrusov: Please add checks like this to event handlers for external messages.
```ts
if (payload.service_id === this.config.serviceId) {
```

I added some in my Localization PR, but these new image-types events I don't
have there, so it might be forgotten.

services/media/service/src/messaging/register-messaging.ts
```diff
@@ -1,39 +1,348 @@
...
+      ...commonMessageHandlers,
+    ],
+    logMapper,
+    {
+      concurrencyStrategy,
+      // Allow the ingest start to be processed a bit longer
+      messageProcessingTimeoutStrategy: (message) =>
+        message.messageType ===
+        MediaServiceMessagingSettings.StartIngest.messageType
+          ? 30_000
```
AxTrusov, line 185: I think StartIngest must be reworked to not use a single transaction to save all
changes, as this is
not enough. E.g. I tried uploading an ingest item with 1000 entities, 1000
images, and 10 videos. And even 60_000 was not enough.

```
[tsc] {
[tsc]   logtime: '2023-12-20T12:21:32.062Z',
[tsc]   loglevel: 'WARN',
[tsc]   retention: 'medium',
[tsc]   context: 'bootstrap',
[tsc]   message: 'An error ocurred while processing the message with id 564d88ea-0667-4c8f-822d-4216386d93f4.',
[tsc]   details: {
[tsc]     tenantId: '00000000-0000-0000-0000-000000000000',
[tsc]     environmentId: '00000000-0000-0000-0000-000000000000'
[tsc]   },
[tsc]   error: {
[tsc]     name: 'Error',
[tsc]     message: 'Could not process the message with ID 564d88ea-0667-4c8f-822d-4216386d93f4 and LSN 6B/C726ED38 within the timeout of 60000 milliseconds. Please consider to use a background worker for long running tasks to not block the message processing.',
[tsc]     stack: 'Error: Could not process the message with ID 564d88ea-0667-4c8f-822d-4216386d93f4 and LSN 6B/C726ED38 within the timeout of 60000 milliseconds. Please consider to use a background worker for long running tasks to not block the message processing.\n' +
[tsc]       '    at Timeout._onTimeout (C:\\Projects\\mosaic-media-template\\node_modules\\pg-transactional-outbox\\dist\\common\\utils.js:31:49)\n' +
[tsc]       '    at listOnTimeout (node:internal/timers:569:17)\n' +
[tsc]       '    at process.processTimers (node:internal/timers:512:7)'
[tsc]   }
[tsc] }
```

AxTrusov: Tried setting timeout to `600_000` and got the following:
```
[tsc] {
[tsc]   logtime: '2023-12-20T12:29:11.579Z',
[tsc]   loglevel: 'ERROR',
[tsc]   retention: 'medium',
[tsc]   context: 'bootstrap',
[tsc]   message: 'Transactional inbox listener DB client error',
[tsc]   details: {
[tsc]     tenantId: '00000000-0000-0000-0000-000000000000',
[tsc]     environmentId: '00000000-0000-0000-0000-000000000000'
[tsc]   },
[tsc]   error: {
[tsc]     name: 'Error',
[tsc]     message: 'Connection terminated unexpectedly',
[tsc]     stack: 'Error: Connection terminated unexpectedly\n' +
[tsc]       '    at Connection.<anonymous> (C:\\Projects\\mosaic-media-template\\node_modules\\pg-transactional-outbox\\node_modules\\pg\\lib\\client.js:132:73)\n' +
[tsc]       '    at Object.onceWrapper (node:events:627:28)\n' +
[tsc]       '    at Connection.emit (node:events:513:28)\n' +
[tsc]       '    at Socket.<anonymous> (C:\\Projects\\mosaic-media-template\\node_modules\\pg-transactional-outbox\\node_modules\\pg\\lib\\connection.js:63:12)\n' +
[tsc]       '    at Socket.emit (node:events:513:28)\n' +
[tsc]       '    at TCP.<anonymous> (node:net:322:12)'
[tsc]   }
[tsc] }
[tsc] {
[tsc]   logtime: '2023-12-20T12:29:11.580Z',
[tsc]   loglevel: 'ERROR',
[tsc]   retention: 'medium',
[tsc]   context: 'bootstrap',
[tsc]   message: 'Transactional inbox listener error',
[tsc]   details: {
[tsc]     tenantId: '00000000-0000-0000-0000-000000000000',
[tsc]     environmentId: '00000000-0000-0000-0000-000000000000'
[tsc]   },
[tsc]   error: {
[tsc]     name: 'Error',
[tsc]     message: 'Connection terminated unexpectedly',
[tsc]     stack: 'Error: Connection terminated unexpectedly\n' +
[tsc]       '    at C:\\Projects\\mosaic-media-template\\node_modules\\pg-transactional-outbox\\node_modules\\pg\\lib\\client.js:526:17\n' +
[tsc]       '    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)'
[tsc]   }
[tsc] }
```

libs/media-messages/schemas/payloads/media/media-service-asyncapi.yml
Zehelein: Yes, the only consumer is in this case the media service so this is fine. Also generally: this is an AsyncAPI "suggestion" on how to name a queue. The defining service of the event makes a suggestion on how the consumer should name its queue.

services/media/service/.env.template
Zehelein: The goal/assumption is that the localization one is finished first. If that will not be the case then yes - those must be added. For now this PR only contains the changes that are required for development to work. Leaving this comment open until we decide what to merge first.

services/media/service/src/messaging/register-messaging.ts
Zehelein: The import ones are the important ones to run in parallel. But also for most of the others they can be run in parallel.

services/media/service/src/ingest/handlers/check-finish-ingest-document-handler.ts
Zehelein: With the change to the polling subscribe I set now the "locked until" to not process the message for the next 5 seconds. So the transaction is immediately committed and the next check is done correctly in 5 seconds. This solution will now not run into issues when the app crashes.

services/media/service/.env.template
Zehelein: Update: the outbox/inbox PR will be completed first but without using logical replication. The polling listener approach is used instead.

scripts/helpers/db-test-helpers.ts
Zehelein: The outbox/inbox PR will not use logical replication but the polling listener approach.

services/media/service/src/publishing/plugins/entity-publishing-endpoints-plugin-factory.ts
Zehelein: Hm - this is (one of the) reasons why we use this plugin :-) The database would still be in a consistent state (only the consumer was not informed of that in a way that he could likely assume). If some customization project would not want to use the plugin they would need to adjust this and other code to use safepoints or refuse to handle more than one mutation per GraphQL call.

services/media/service/src/messaging/register-messaging.ts
Zehelein: Yes, I improved it a bit but generally a very big ingest would need to be split up. I will create a task that includes reworking the ingest messaging: all messages staying inside of the media service can be directly added to the inbox and do not need to go through RabbitMQ now anymore.

services/media/service/migrations/committed/000022-transactional-outbox-inbox.sql
```diff
@@ -0,0 +1,235 @@
...
+
+AS $BODY$
+DECLARE 
+  loop_row app_hidden.outbox%ROWTYPE;
+  message_row app_hidden.outbox%ROWTYPE;
+  ids uuid[] := '{}';
+BEGIN
+
+  IF max_size < 1 THEN
+    RAISE EXCEPTION 'The max_size for the next messages batch must be at least one.' using errcode = 'MAXNR';
```
AxTrusov, line 49: If this error is thrown - it will currently be mapped to the UNHANDLED_ERROR. If
this is not the intention - `MAXNR` should be mapped in the `services\media\service\src\common\errors\media-pg-error-mapper.ts`

services/media/service/migrations/committed/000022-transactional-outbox-inbox.sql
```diff
@@ -0,0 +1,235 @@
...
+  END IF;
+
+  -- get (only) the oldest message of every segment but only return it if it is not locked
+  FOR loop_row IN
+    SELECT * FROM app_hidden.outbox m WHERE m.id in (SELECT DISTINCT ON (segment) id
+      FROM app_hidden.outbox
+      WHERE processed_at IS NULL AND abandoned_at IS NULL
+      ORDER BY segment, created_at) order by created_at
```
AxTrusov, line 57: Very minor, but would it make sense to change this to `order by created_at ASC` to make it more explicit?

services/media/service/migrations/committed/000022-transactional-outbox-inbox.sql
```diff
@@ -0,0 +1,235 @@
...
+        RETURNING *;
+
+  END IF;
+END;
+$BODY$;
+
+CREATE INDEX outbox_segment_idx ON app_hidden.outbox (segment);
+CREATE INDEX outbox_created_at_idx ON app_hidden.outbox (created_at);
+CREATE INDEX outbox_processed_at_idx ON app_hidden.outbox (processed_at);
+CREATE INDEX outbox_abandoned_at_idx ON app_hidden.outbox (abandoned_at);
```
AxTrusov: Do we need to also add an index for `locked_until` (for both outbox and inbox tables), since its used for filtering
messages with parallel concurency?

services/media/service/src/domains/common/handlers/delete-entity-command-handler.ts
```diff
@@ -1,67 +1,68 @@
...
 } from 'media-messages';
-import { deletes, IsolationLevel } from 'zapatos/db';
+import { ClientBase } from 'pg';
+import { deletes } from 'zapatos/db';
 import { Table } from 'zapatos/schema';
 import { Config } from '../../../common';
-import { MediaGuardedMessageHandler } from '../../../messaging';
+import { MediaGuardedTransactionalInboxMessageHandler } from '../../../messaging';
 
-export class DeleteEntityCommandHandler extends MediaGuardedMessageHandler<DeleteEntityCommand> {
```
AxTrusov, line 14: Would also change the file name of this file and related test file to not includ `command`

services/media/service/migrations/committed/000022-transactional-outbox-inbox.sql
Zehelein: We use the mapping for errors that (may) show up to end users like in the GraphQL API. The outbox/inbox logic is purely internal and should rather log the technical error.

services/media/service/migrations/committed/000022-transactional-outbox-inbox.sql
Zehelein: Omitting the "ASC" is in my experience the most often used approach - adding only DESC where needed.

services/media/service/migrations/committed/000022-transactional-outbox-inbox.sql
Zehelein: Good point - added it and switched to the idempotent `ax_define` variant.

services/media/service/src/ingest/handlers/check-finish-ingest-document-handler.ts
```diff
@@ -1,89 +1,79 @@
...
-  status: IngestItemStatusEnum;
-  count: number;
-}
+import { MediaGuardedTransactionalInboxMessageHandler } from '../../messaging';
+import { getFutureIsoDateInMilliseconds } from '../utils';
 
-export class CheckFinishIngestDocumentHandler extends MediaGuardedMessageHandler<CheckFinishIngestDocumentCommand> {
+export class CheckFinishIngestDocumentHandler extends MediaGuardedTransactionalInboxMessageHandler<
+  CheckFinishIngestDocumentCommand,
+  Config
```
AxTrusov: I think it should be possible to remove this `Config` generic parameter by changing the `MediaGuardedTransactionalInboxMessageHandler` in approximately the following way?
image

services/media/service/src/messaging/transactional-outbox-inbox/media-guard-transactional-message-handler.ts
```diff
@@ -0,0 +1,56 @@
...
+  GuardedConfig,
+  GuardedTransactionalInboxMessageHandler,
+} from '@axinom/mosaic-id-guard';
+import { MessagingSettings } from '@axinom/mosaic-message-bus-abstractions';
+import { getMappedError, Logger } from '@axinom/mosaic-service-common';
+import { ClientBase } from 'pg';
+import { Config } from '../../common';
+import { PermissionKey } from '../../domains/permission-definition';
+
+export type MediaGuardedConfig = GuardedConfig & {
```
AxTrusov: This is now not used anywhere and can be dropped.

AxTrusov (approved):

PR #253, opened 2024-03-04
[AB#42975] logical replication approach dropped in favor of db triggers and message polling

AxTrusov:
# Pull Request

## Prerequisites

      releases)

## Description

Previously, we used logical replication to detect database changes related to localizable tables. Now, each table is explicitly handled by having AFTER INSERT/AFTER UPDATE/AFTER DELETE triggers (using dedicated db helper functions). Each trigger inserts a relevant entry into the `inbox` table, which is processed by the `pg-transactional-outbox` message polling mechanism to send messages to the localization service in a similar way it was done with logical replication.

## Testing notes

General testing on localization-related functionality shall be done, e.g. Create, update, delete, assign cover, change cover, unassign cover for movie/tvshow/season/episode/collection. Including Ingest and Publishing.


services/media/service/migrations/committed/000023-localization-changes.sql
```diff
@@ -0,0 +1,176 @@
...
+
+INSERT INTO app_public.snapshot_validation_issue_context (value, description)
+VALUES ('LOCALIZATION', 'Localization')
+ON CONFLICT (value) DO NOTHING;
+
+INSERT INTO app_public.ingest_item_step_type (value, description)
+VALUES ('LOCALIZATIONS', 'Localizations')
+ON CONFLICT (value) DO NOTHING;
+
+ALTER TABLE app_public.movies ADD COLUMN IF NOT EXISTS ingest_correlation_id INT;
```
Zehelein, line 29: Minor: maybe you can add a comment here on why such a correlation field is needed.

services/media/service/migrations/committed/000023-localization-changes.sql
```diff
@@ -0,0 +1,176 @@
...
+*/
+CREATE OR REPLACE FUNCTION app_hidden.is_localization_enabled()
+  RETURNS boolean AS
+  $f$SELECT FALSE $f$ LANGUAGE sql IMMUTABLE;
+
+/*
+Generates an AFTER INSERT, AFTER UPDATE, and AFTER DELETE trigger for the
+specified table. Triggers insert an entry into the app_hidden.inbox table
+whenever a change that is relevant for the localization is done on the table.
+Localizable and Required field parameters are used to determine in entry should
```
Zehelein: I do not understand this comment. Should they determine if an entry should be inserted into the inbox if localizable and required fields were inserted/updated?

services/media/service/src/common/db/set-is-localization-enabled-db-function.ts
```diff
@@ -0,0 +1,18 @@
+import { OwnerPgPool } from '@axinom/mosaic-db-common';
+import { raw, sql } from 'zapatos/db';
+
+/**
+ * Overrides the `app_hidden.is_localization_enabled()` function by making it
+ * return the value of `IS_LOCALIZATION_ENABLED` variable, enabling or disabling
+ * the localizable entity triggers. Call this on every service startup to align
+ * with the latest config value.
```
Zehelein, line 8: Question: should we add a hint here if they would surely know to use or not use localization that they can just do this as part of a DB migration and not call this function on every service startup?

services/media/service/src/domains/common/handlers/localizable-media-transactional-inbox-message-handler.ts
```diff
@@ -0,0 +1,66 @@
...
+  override async handleMessage(
+    message: TypedTransactionalMessage<T>,
+    loginClient: ClientBase,
+  ): Promise<void> {
+    const data = await this.getLocalizationCommandData(message, loginClient);
+    if (!data) {
+      return;
+    }
+
+    const accessToken = await requestServiceAccountToken(this.config);
```
Zehelein, line 44: Can we cache this - best in some service-global scope (until the token expires/a few minutes befor that)? Right now this would do a synchronous request on the ID service for every localization related change.

services/media/service/src/domains/collections/localization/localizable-collection-db-messaging-settings.ts
```diff
@@ -0,0 +1,66 @@
...
+      '',
+      'event',
+      'collection-image',
+    );
+
+  private constructor(
+    public readonly messageType: string,
+    public readonly queue: string,
+    public readonly routingKey: string,
+    public readonly action: 'command' | 'event',
```
Zehelein, line 52: As the routing key is always string empty you could also directly have it manually set to string empty or use `public readonly routingKey: string = '',` (as the last parameter)

services/media/service/migrations/committed/000023-localization-changes.sql
```diff
@@ -0,0 +1,176 @@
...
+
+`tests/localization` folder contains tests for these triggers, including mapping
+of messaging-related columns to the MessagingSettings that would be used for
+pulling the inbox entries and executing relevant message handler logic. 
+*/
+CREATE OR REPLACE FUNCTION app_hidden.create_localizable_entity_triggers(aggregateId text, tableName text, entityType text, localizable_fields text, required_fields text) RETURNS void
+    LANGUAGE plpgsql
+    AS $$
+BEGIN
+    EXECUTE 'CREATE OR REPLACE FUNCTION app_hidden.localizable_' || entityType || '_insert() RETURNS TRIGGER AS $body$ ' ||
```
Zehelein: If easy: could you adjust this function in a way that it uses new lines? In the generated SQL all those functions are on a single line.

services/media/service/src/ingest/handlers/localize-entity-failed-handler.ts
```diff
@@ -1,60 +1,65 @@
...
+      aggregateId,
+    }: TypedTransactionalMessage<LocalizeEntityFailedEvent>,
+    loginClient: ClientBase,
   ): Promise<void> {
-    if (content.service_id !== this.config.serviceId) {
+    if (
+      !checkIsIngestEvent(metadata, this.logger, id, aggregateId) ||
+      payload.service_id !== this.config.serviceId
+    ) {
       // skipping events for entity types from different services
```
Zehelein: the comment could be adjusted to include that we also skip non-ingest events (also in the other ingest handlers).

services/media/service/src/tests/localization/localizable-collection-triggers.db.spec.ts
```diff
@@ -0,0 +1,489 @@
...
+            id: collection.id,
+          },
+        },
+      ]);
+    });
+  });
+
+  describe('collections_images', () => {
+    it('Collection COVER inserted -> inbox entry added', async () => {
+      // TEASER is expected to be skipped in the message handler
```
Zehelein: that comment can be removed in this unit test

services/media/service/src/domains/common/handlers/localizable-media-transactional-inbox-message-handler.ts
AxTrusov: `requestServiceAccountToken` already contains caching within itself.

Zehelein (approved):

PR #291, opened 2024-04-23
[AB#43238] start ingest now processes items in batches

AxTrusov:
# Pull Request

## Prerequisites

      releases)

## Description

- Previously, ingesting large ingest files (500+ items) would cause ingest to fail without processing any.
- The cause was trying to insert too much data in the same database transaction.
- Current approach switched to processing items in batches in separate transactions
- To accommodate this, several loosely-related adjustments were made   
     - Service-account token longevity extended from 1 hour to 7 days (and cache refresh still happens after one hour to account for possible permission changes. )
     - Messages retries number increased from 5 to 10 (to accommodate for 5000+ items ingests) only for ingest-related messages
        - Further optimizations expected regarding this, so the retries number might go back down.
     - Timeout for `StartIngest` command processing increased to 10 minutes.
     
## Testing notes

- Try ingesting a file with 500+ items with images and localizations.
    - Suggesting to not use videos for testing to avoid loading the encoder.
- Expected result - it should succeed.

Expect a longer ingest run. E.g. ~30 minutes to process 2500 items, each of which has 3 localizations and 1 image.


services/media/service/migrations/committed/000025-ingest-document-started-count-added.sql
```diff
@@ -0,0 +1,6 @@
+--! Previous: sha1:d7c481931580f6acf907cf543994742775ae710c
+--! Hash: sha1:9e58d9212f8204cc42f4e3b3afd0abe7d343bd12
+--! Message: ingest-document-started-count-added
+
+ALTER TABLE app_public.ingest_documents ADD COLUMN IF NOT EXISTS started_count INT NOT NULL DEFAULT 0;
+UPDATE app_public.ingest_documents SET started_count = items_count WHERE started_count = 0;
```
Zehelein, line 6: Should this rather be a count from the ingest_items associated with this ingest document?

services/media/service/src/common/utils/token-utils.ts
```diff
@@ -5,8 +5,13 @@ import {
 import NodeCache from 'node-cache';
 import { Config } from '../config';
 
-// By default, service account token is valid for 1 hour
-const cache = new NodeCache({ stdTTL: 55 * 60 });
+const tokenExpirationInSeconds = 7 * 24 * 60 * 60; // 7 days
```
Zehelein, line 8: Thinking again: could we maybe change this to cache per ingest document? My main concern is about permission updates - if I add or remove permissions from the service account you may need to wait for a week until they work. If we would use a caching key like "ingest-<ingest-doc-id>" we would still only generate the token once per ingest (or in distributed environments once per service) but could avoid caching for such a long time for everything else.

We could then likely just ignore to optimize for refreshing the token as an ingest really should not take 7 days :-)

services/media/service/src/ingest/handlers/start-ingest-handler.ts
```diff
@@ -50,30 +71,92 @@ export class StartIngestHandler extends MediaGuardedTransactionalInboxMessageHan
...
+    // Establish the order by which items of specific type would be processed,
+    // based on the order of entity processors.
+    const orderedTypes: Dict<number> = this.entityProcessors.reduce<
+      Dict<number>
+    >((result, value, index) => {
+      result[value.type] = index;
+      return result;
+    }, {});
+    const notStartedItems: IngestItem[] = document.items.sort(
+      (a, b) => orderedTypes[a.type] - orderedTypes[b.type],
```
Zehelein: According to https://stackoverflow.com/questions/37020643/javascript-sorting-strange-behaviour-any-idea "The sort is not necessarily stable (that is, elements that compare equal do not necessarily remain in their original order)". So a second comparison is required here - first sort on the ingest type and then on the natural index within the document. So if there are two movies that the sorting function guarantees that the first movie in the ingest document always comes before a later movie in the ingest document.

services/media/service/migrations/committed/000025-ingest-document-started-count-added.sql
AxTrusov: It is. `items_count` represents how many ingest items are associated with the document

services/media/service/src/ingest/handlers/start-ingest-handler.ts
AxTrusov: While my tests did produce consistent ordering results, It would be nice to have some explicit sorting guarantee. I think the second sort value should be `external_id`. Will try to implement this.

services/media/service/src/common/utils/token-utils.ts
AxTrusov: If we want to cache it by ingest document - we would have to pass ingest document id as part of `ingest_correlation_id` property, in addition to the current value of `Ingest Item ID`. Because the place where we call the getServiceToken function is the message handler after the db trigger. Do we want to go this way?

AxTrusov: For now, token is valid for 7 days and refreshes every hour. Expecting a more sophisticated cache invalidation approach in the future.

Zehelein (approved):

PR #293, opened 2024-04-29
[AB#43320] check-finished-ingest-item message dropped

AxTrusov:
# Pull Request

## Prerequisites

      releases)

## Description

- `CheckFinishIngestItem` command dropped in favor of setting ingest item step state directly and making `CheckFinishIngestDocument` update ingest item states based on step states in a single update statement.
- `app_hidden.next_inbox_messages` updated to always try to load 1 `ingest-document` related message (`StartIngest` or `CheckFinishIngestDocument`) as a first priority message.
   - This ensures ingest is getting started faster and states are updated frequently (and not a couple of times at the end of ingest)

## Testing notes

Make sure ingest works, same as before (will probably finish a faster, but might only be noticeable with larger ingests)


services/media/service/src/ingest/handlers/check-finish-ingest-document-handler.ts
```diff
@@ -43,10 +43,30 @@ export class CheckFinishIngestDocumentHandler extends MediaGuardedTransactionalI
     }: TypedTransactionalMessage<CheckFinishIngestDocumentCommand>,
     ownerClient: ClientBase,
   ): Promise<void> {
+    const docId = param(ingest_document_id);
+    await sql`WITH updated AS (
```
AxTrusov, line 47: Checked this query on a ~4500 ingest-items document. It finishes in ~80-90ms. So, I think peformance-wise it should be acceptable.

services/media/service/migrations/committed/000026-next-inbox-message-ingest-priority-added.sql
```diff
@@ -0,0 +1,112 @@
...
+        NULL;
+    WHEN serialization_failure THEN
+        NULL;
+    WHEN OTHERS THEN
+        RAISE;
+  END;
+
+  -- get (only) the oldest message of every segment but only return it if it is not locked
+  FOR loop_row IN
+    SELECT * FROM app_hidden.inbox m WHERE m.id in (SELECT DISTINCT ON (segment) id
```
Zehelein: Can you try combining the query in line 25 with this one e.g. via union? Something like this (not tested):

```
(SELECT * FROM app_hidden.inbox m WHERE m.id in (SELECT DISTINCT ON (segment) id
  FROM app_hidden.inbox
  WHERE processed_at IS NULL AND abandoned_at IS NULL AND aggregate_type = 'ingest-document'
  ORDER BY segment, created_at) order by created_at limit 1)
  
UNION

(SELECT * FROM app_hidden.inbox m WHERE m.id in (SELECT DISTINCT ON (segment) id
  FROM app_hidden.inbox
  WHERE processed_at IS NULL AND abandoned_at IS NULL
  ORDER BY segment, created_at) order by created_at)
```

services/media/service/src/ingest/handlers/check-finish-ingest-document-handler.ts
Zehelein: Can you check that there is a combined index on document_id plus status on the ingest item?

services/media/service/src/ingest/handlers/image-succeeded-handler.ts
```diff
@@ -106,19 +86,16 @@ export abstract class ImageSucceededHandler<
...
+    await update(
+      'ingest_item_steps',
       {
-        ingest_item_step_id: messageContext.ingestItemStepId,
-        ingest_item_id: messageContext.ingestItemId,
-        error_message: getIngestErrorMessage(
+        status: 'ERROR',
+        response_message: getIngestErrorMessage(
           error,
           'An unexpected error occurred while trying to update image relations.',
```
Zehelein, line 95: side note: could we maybe update this to something like 

* The image [was correctly imported | already existed] but there was an error adding that image to the entity.

services/media/service/src/ingest/handlers/video-succeeded-handler.db.spec.ts
```diff
@@ -95,7 +84,6 @@ describe('VideoSucceededHandler', () => {
 
   afterEach(async () => {
     await ctx.truncate('ingest_documents');
-    payloads = [];
   });
 
   afterAll(async () => {
```
Zehelein, line 89: onMessage and onMessageFailure descriptions could be renamed to handleMessage and handleErrorMessage.

services/media/service/src/ingest/handlers/image-succeeded-handler.ts
AxTrusov: Will update in context of other upcoming error-handling-related tasks.

services/media/service/migrations/committed/000026-next-inbox-message-ingest-priority-added.sql
AxTrusov: As discussed, no need to do this. Instead, appropriate messages now have a priority segment assigned to them.

services/media/service/src/ingest/handlers/check-finish-ingest-document-handler.ts
AxTrusov: There was non, only individual indexes. Added.

Zehelein (approved):

PR #295, opened 2024-05-02
[AB#43917] [AB#42728] ingest error handling improved

AxTrusov:
# Pull Request

## Prerequisites

      releases)

## Description

- Custom validation will not be executed on ingest document upload, if schema validation fails.
	- Otherwise - custom validation can fail, e.g. because there is an object isntead of an expected array.
- Fallback error messages are now more detailed.
- Mosaic packages updated to make sure errors are mapped better

## Testing Notes

- Try uploading some ingest document, where `images` is an object, e.g. `{ "type": "COVER", "path": "xzy" }`, instead of an array of objects.
	- Expected error: `Ingest Document validation has failed.` with one of the validation errors being `JSON path \"document/items/0/data/images\" should be array,null (line: 36, column: 19)`
- Try uploading some ingrst document, where some invalid genre is specified (one that is not define in the management system), for movie, tvhsow, season, and episode.
	- Expected error: `Metadata update has failed because following genres do not exist: AAction`
	- Localization step will also fail with a message `Unable to start the localization step because the metadata update has failed.` (instead of being stuck in IN_PROGRESS state)


services/media/service/src/ingest/handlers/video-succeeded-handler.ts
```diff
@@ -72,6 +72,19 @@ export abstract class VideoSucceededHandler<
     ).run(ownerClient);
   }
 
+  public override mapError(error: unknown): Error {
+    const fallbackMessage =
+      this.messagingSettings.messageType ===
+      VideoServiceMultiTenantMessagingSettings.EnsureVideoExistsAlreadyExisted
```
Zehelein: Here and in the other similar handler: as you are anyways using abstract classes I would suggest to not do such if statement here but call an abstract method/property where the child class would then provide that text.

services/media/service/src/ingest/plugins/start-ingest-endpoint-plugin.ts
```diff
@@ -118,16 +118,26 @@ export const StartIngestEndpointPlugin = makeExtendSchemaPlugin((build) => {
...
               );
               throw new MosaicError({
                 ...CommonErrors.IngestValidationError,
                 details: {
-                  validationErrors: [...schemaErrors, ...customErrors],
+                  validationErrors: [...schemaErrors],
+                },
+              });
+            }
+            // Skip custom errors if schema validation failed, since incorrect
```
Zehelein: Not sure about the place for this comment - would it logically make more sense in line 126 where you throw the error? I read this comment and assumed it was a comment for line 135/136 and was wondering how that is skipped :-)

Zehelein (approved):

PR #315, opened 2024-06-06
[AB#44512]: added aggregate ID descriptions for messages

Zehelein:
# Pull Request

## Prerequisites

      releases)

## Description

The messaging settings contain now a description of the command/event as JSDoc string. This explains if the message is of type event or command, the message type, the aggregate type, and now also the aggregate ID field or UNKNOWN_ID if there is no single ID (e.g. for bulk operations).


services/media/service/src/domains/movies/plugins/all-movie-plugins.ts
```diff
@@ -22,7 +22,7 @@ export const AllMoviePlugins = makePluginByCombiningPlugins(
     'MoviesLicenseFilter',
   ),
   EntityListPublishingEndpointsPluginFactory({
-    id: 1,
+    id: -1,
```
AxTrusov: It seems like this change (and the one for tvshows) has broken some tests which
were expecting the ID to be 1. Is there an explicit need to set this to -1?
Would 0 work? (I'm fine with changing it, but not fond of the minus)

AxTrusov (approved):
Looks good as long as unit tests failure is resolved. (see my other comment)

services/media/service/src/domains/movies/plugins/all-movie-plugins.ts
Zehelein: Yes - started to change it to zero already on Friday but had some meetings - will finish that today.

PR #317, opened 2024-06-10
[AB#41605] localizations view optimized in catalog service

AxTrusov:
# Pull Request

## Prerequisites

      releases)
- [x] potential **release notes** to the PR description added

## Release notes

- Localization views adjusted to be more performant and easily trigger indexes. 
- As a result - the view became simpler and will now just return rows for the specified locale, but will not perform any fallback logic to the default locale.
- To preserve the previous logic - the fallback to the default locale now happens on the code level (and not database level), by checking if specified locale exists, and if it does not - use default locale.
- It is now expected that all entities in the catalog service should have localizations for the same locales to avoid inconsistent request results. (was already ensured by the media service publish validation)
- In case a new locale is introduced to be supported - you might want to consider adding an explicit database migration to add fallback rows for new locale until it is properly published from the media service (e.g. using provided `app_private.add_default_placeholder_localizations` function)
- As an example change - `title` property can now be used in filters and ordering and has corresponding indexes added to support performant querying.

## Description

In addition to the Release notes:

- `locales` endpoint available to query a list of available locales
- An in-memory list of active locales is now being maintained during the service runtime.
   - it is initialized on service startup and updated every time a *Published event is received for the localizable entity.
- Locales header filter is now case-insensitive (e.g. filtering by `de-de` and `de-DE` would return the same result and not fall back to default locale. 
- Listen/Notify approach implemented to update in-memory locales array whenever new locales are inserted. This is done mainly horizontal scaling cases, e.g.
   - one instance of catalog service receives a Published message and updates the locales table
   - Notification is emitted
   - Every instance of the catalog service receives it and updates its in-memory locales array.

## Testing notes

- Have 2 movies with some non-default localizations, publish them, check in the catalog service that they can be requested without locale header and with locale header, and that values for appropriate locale is returned.
   - In Portal Admin - deactivate some locale
   - re-publish one of the above movies - this will update the in-memory locales to no longer include deactivated locale
   - query both re-published and second movie for deactivated locale - both would return default fallback locale
   - Expectation for "deactivating old locale" in prod environment scenario - re-publishing of all movies to drop obsolete localizations or explicit database migration to drop no longer supported localizations

- Have 2 movies with some non-default localizations, publish them, check in the catalog service that they can be requested without locale header and with locale header, and that values for appropriate locale is returned.
   - In Portal Admin - add and activate a new locale
   - For one of the movies - fill in localizations for the new locale and re-publish it.
   - Requesting re-published movie for new locale would return it appropriately
   - Requesting not re-published movie (that does not have a new locale) - would return null (if a single movie is requested), or not be returned (if a `movies` query is used)
   - Expectation for "activating a new locale" in prod environment scenario -  either re-publish all entities or add an explicit database migration to add fallback localizations rows for new locale

- After one of the cases above (either works) - check the logs - there should be a single log entry with message `In-memory locales successfully (re)loaded.`
   - Ask SE Team to scale the catalog service horizontally to have 2 instances (pods).
   - After that is done - retry a case by either activating or de-activating a locale and re-publishing the movie.
   - Check the logs again - it should have 2 new `In-memory locales successfully (re)loaded.` logs, one for each running catalog service instance.

### DIFF
--- a/services/catalog/service/src/common/db/get-mosaic-locale-setting.spec.ts
+++ b/services/catalog/service/src/common/db/get-mosaic-locale-setting.spec.ts
@@ -26,11 +26,11 @@ describe('getMosaicLocaleSetting', () => {
 
   it.each([null, undefined, '', ' ', 'my-locale', 'de-DE', ['asd']])(
     'No supported locales and any header %p specified --> setting with default locale returned',
-    async (emptyValue) => {
+    async (value) => {
       // Arrange
       const req = mockRequest({
         headers: {
-          [MOSAIC_LOCALE_HEADER_KEY]: emptyValue,
+          [MOSAIC_LOCALE_HEADER_KEY]: value,
         },
       });
 
--- a/services/catalog/service/src/common/db/get-mosaic-locale-setting.spec.ts
+++ b/services/catalog/service/src/common/db/get-mosaic-locale-setting.spec.ts
@@ -1,0 +1,109 @@
+import { mockRequest } from 'mock-req-res';
+import {
+  DEFAULT_LOCALE_TAG,
+  MOSAIC_LOCALE_HEADER_KEY,
+  MOSAIC_LOCALE_PG_KEY,
+} from '../constants';
+import { getMosaicLocaleSetting } from './get-mosaic-locale-setting';
+import * as m from './in-memory-locales';
+
+describe('getMosaicLocaleSetting', () => {
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+  it('No supported locales and no header specified -> setting with selected default returned', async () => {
+    // Arrange
+    const req = mockRequest({ headers: {} });
+
+    // Act
+    const setting = getMosaicLocaleSetting(req);
+
+    // Assert
+    expect(setting).toEqual({
+      [MOSAIC_LOCALE_PG_KEY]: DEFAULT_LOCALE_TAG,
+    });
+  });
+
+  it.each([null, undefined, '', ' ', 'my-locale', 'de-DE', ['asd']])(
+    'No supported locales and any header %p specified --> setting with default locale returned',
+    async (emptyValue) => {
+      // Arrange
+      const req = mockRequest({
+        headers: {
+          [MOSAIC_LOCALE_HEADER_KEY]: emptyValue,
+        },
+      });
+
+      // Act
+      const setting = getMosaicLocaleSetting(req);
+
+      // Assert
+      expect(setting).toEqual({
+        [MOSAIC_LOCALE_PG_KEY]: DEFAULT_LOCALE_TAG,
+      });
+    },
+  );
+
+  it.each(['de-DE', 'de-de', 'DE-DE', 'DE-de'])(
+    'Supported locale exists and matching header %p specified --> setting with selected locale returned',
+    async (header) => {
+      // Arrange
+      const locale = 'de-DE';
+      jest.spyOn(m, 'getInMemoryLocales').mockImplementation(() => [locale]);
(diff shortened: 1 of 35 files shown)


services/catalog/service/migrations/committed/000015-localizations-view-approach-adjusted.sql
```diff
@@ -0,0 +1,237 @@
...
+*/
+CREATE OR REPLACE FUNCTION app_private.define_localization_view(localizableFields text[], tableName text, localizationsTableName text, fkColumn text) RETURNS void
+    LANGUAGE plpgsql
+    AS $$
+DECLARE
+  localizableFieldsSelect text = 'l.' || array_to_string(localizableFields, ', l.');
+BEGIN
+  EXECUTE 'DROP VIEW IF EXISTS app_public.' || tableName || '_view CASCADE;';
+  EXECUTE 'CREATE VIEW app_public.' || tableName || '_view AS ' ||
+          'SELECT p.*, ' || localizableFieldsSelect || ' FROM app_public.' || localizationsTableName || ' as l ' ||
```
Zehelein: Maybe we could improve that one in a similar way like further down: we can select all rows from the table and exclude the known id/is_default/... so only the actual localizable ones remain. Then we would not need to provide them as input.

services/catalog/service/src/common/db/get-mosaic-locale-setting.spec.ts
```diff
@@ -0,0 +1,109 @@
...
+    const setting = getMosaicLocaleSetting(req);
+
+    // Assert
+    expect(setting).toEqual({
+      [MOSAIC_LOCALE_PG_KEY]: DEFAULT_LOCALE_TAG,
+    });
+  });
+
+  it.each([null, undefined, '', ' ', 'my-locale', 'de-DE', ['asd']])(
+    'No supported locales and any header %p specified --> setting with default locale returned',
```
Zehelein, line 28: Line 29 is about "empty" values and line 28 about "any header" - could you please clarify/reword?

services/catalog/service/src/common/db/in-memory-locales.ts
```diff
@@ -0,0 +1,74 @@
...
+import { all, deletes, insert, Queryable, select, SQL, sql } from 'zapatos/db';
+
+let inMemoryLocales: string[] = [];
+
+/**
+ * Returns a list of locales that currently exist in the system.
+ * If an API request is made for a specific locale and it is not in this array -
+ * default locale would be used instead.
+ */
+export const getInMemoryLocales = (): string[] => inMemoryLocales;
```
Zehelein, line 19: Thinking of it again: this would be a problem if the Catalog Service runs in multiple instances. Maybe we can add listen/notify (https://node-postgres.com/apis/client#notification) and add triggers for insert/delete and updates (only ) in that locales table? 

And just in case if the notify fails for whatever reason we could do a background call every 5 minutes or so to update the inmemory cache?

services/catalog/service/src/common/db/in-memory-locales.ts
```diff
@@ -52,23 +59,94 @@ export const syncInMemoryLocales = async (
...
+      error,
+      `Listener database connection error occurred. Attempting to reconnect. (${currentRetry}/${maxRetryCount})`,
+    );
+    await sleep(currentRetry * 1000);
+    await startLocalesInsertedListener(config, logger);
+  };
+
+  try {
+    closeActiveClient();
+    // TODO: add testing notes for scaling + pgBouncer
```
Zehelein: Maybe we do not need that hint now any more for pgBouncer?

Zehelein (approved):

PR #357, opened 2024-08-22
[AB#45596] in-memory locales now stored as objects and default locale fallback considers is_default value

AxTrusov:
# Pull Request

## Prerequisites

      releases)

## Testing notes

- Make sure you have the `Default Locale Language Tag` in Admin Portal set to the default value `default`
- Publish some movie
- Request all movies from the Catalog GraphQL API without specifying the `mosaic-locale` header.
    - Expected: published movie is returned
- Change the `Default Locale Language Tag` in Admin Portal to something else, e.g. `en-US`
- Re-publish previously published movie
- Request all movies from the Catalog GraphQL API without specifying the `mosaic-locale` header.
    - Expected: published movie is returned
    - Previous behavior: Movie was not returned, because `is_default` property was not considered when performing a fallback to the default locale.
    
Control case:
- Try requesting movies from Catalog GraphQL API with `mosaic-locale`, requesting both for default and not default locales. Expected that movies are returned for specified locales with matching localized values.

### DIFF
--- a/services/catalog/service/src/common/db/in-memory-locales.ts
+++ b/services/catalog/service/src/common/db/in-memory-locales.ts
@@ -6,17 +6,34 @@ import {
 } from '@axinom/mosaic-service-common';
 import { Client } from 'pg';
 import { all, deletes, insert, Queryable, select } from 'zapatos/db';
+import { locales } from 'zapatos/schema';
 import { Config } from '../config';
-import { MOSAIC_LOCALE_NOTIFY } from '../constants';
+import { DEFAULT_LOCALE_TAG, MOSAIC_LOCALE_NOTIFY } from '../constants';
 
-let inMemoryLocales: string[] = [];
+let inMemoryLocales: locales.JSONSelectable[] = [];
 
 /**
- * Returns a list of locales that currently exist in the system.
- * If an API request is made for a specific locale and it is not in this array -
- * default locale would be used instead.
+ * Returns a language tag for a requested locale tag from the in-memory array of
+ * locales currently registered within the service.
+ * If not found - locale that is marked as default will be returned.
+ * If such locale is not found - fallback to hardcoded `DEFAULT_LOCALE_TAG` is
+ * used.
  */
-export const getInMemoryLocales = (): string[] => inMemoryLocales;
+export const getInMemoryLocale = (locale: string): string => {
+  const found = inMemoryLocales.find(
+    (l) => l.locale.toLowerCase() === locale.toLowerCase(),
+  );
+  if (found) {
+    return found.locale;
+  }
+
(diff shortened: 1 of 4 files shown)


services/catalog/service/src/common/db/in-memory-locales.ts
```diff
@@ -6,17 +6,34 @@ import {
...
-export const getInMemoryLocales = (): string[] => inMemoryLocales;
+export const getInMemoryLocale = (locale: string): string => {
+  const found = inMemoryLocales.find(
+    (l) => l.locale.toLowerCase() === locale.toLowerCase(),
+  );
+  if (found) {
+    return found.locale;
+  }
+
+  const def = inMemoryLocales.find((l) => l.is_default);
```
yasithA: minor cosmetics - shall we rename this variable to something like `defaultLocale` or similar  just so it's clear that this is the default locale at the first glance?

services/catalog/service/src/common/db/in-memory-locales.ts
yasithA, line 1: I can't comment this on the line because it's not modified - but any chance we could use `locales.JSONSelectable` as the type for `input` in line 47? Or is the language tag is different to a locale?

yasithA (approved):

services/catalog/service/src/common/db/in-memory-locales.ts
AxTrusov: I was considering it, but I don't like that below it will look like `defaultLocale.locale` :)
And `default` is not a valid variable name... I think `is_default` in find is a good-enough indication

services/catalog/service/src/common/db/in-memory-locales.ts
AxTrusov: They are the same. But, the main use case for `syncInMemoryLocales` looks like this:
```ts
await syncInMemoryLocales(payload.localizations, txnClient);
```

Current property names match the metadata received with EntityPublished events, and I'd rather not change that format without an explicit need. And doing the mapping in each event handler also seems wrong. So, this seemed like the less evil option.

services/catalog/service/src/common/db/in-memory-locales.ts
yasithA: I was thinking about `defaultLocale.locale` as well. First thought maybe `defaultSetting` would be a better fit. But then I saw we're actually having some Postgres settings set for locales as well so thought against it.

It's just when I saw `def`, the first that came to my mind was `definition` and took a second to realize it means default. xD

I'll leave it to you to decide if we really want to rename or not. Think we can live with `is_default`

AxTrusov: Decided to go with `foundDefault`

PR #494, opened 2025-11-25
[AB#52395] fix: :bug: handle deferred constraints separately for `collection_relations`

EmperorRXF:
# Pull Request

## Prerequisites

- [X] The PR is targeting the `dev` branch

## Description

Added custom handling logic for the `BulkEditItemChangeHandler` to care for the table `collection_relations` since it uses a `DEFERRABLE INITIALLY DEFERRED` syntax that is not supported together with `ON CONFLICT` syntax that the Bulk Edit plugin uses for `ADD_RELATED_ENTITY`

### DIFF
--- a/services/media/service/src/domains/common/handlers/bulk-edit-item-change-handler.ts
+++ b/services/media/service/src/domains/common/handlers/bulk-edit-item-change-handler.ts
@@ -44,14 +44,65 @@ export class BulkEditItemChangeHandler extends MediaGuardedTransactionalInboxMes
   ): Promise<void> {
     this.logger.debug({ details: { ...message.payload } });
 
-    const handled = await this.handleImageTableAddRelated(
+    const imageHandled = await this.handleImageTableAddRelated(
       message.payload,
       envOwnerClient,
     );
 
-    if (!handled) {
+    if (imageHandled) {
+      return;
+    }
+
+    const deferrableHandled = await this.handleDeferrableConstraintTables(
+      message.payload,
+      envOwnerClient,
+    );
+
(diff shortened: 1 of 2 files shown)


services/media/service/src/domains/common/handlers/bulk-edit-item-change-handler.ts
```diff
@@ -44,16 +44,67 @@ export class BulkEditItemChangeHandler extends MediaGuardedTransactionalInboxMes
...
     }
   }
 
+  private async handleDeferrableConstraintTables(
+    payload: PerformItemChangeCommand,
+    envOwnerClient: ClientBase,
+  ): Promise<boolean> {
+    // Handle collection_relations which has a DEFERRABLE sort_order constraint
+    // that conflicts with the `ON CONFLICT` clauses in the base `handlePerformItemChangeCommand` handler.
+    if (payload.table_name !== 'collection_relations') {
```
yasithA, line 72: Can we have this in an array, that is easily updateable so if more tables are created with DEFERRABLE constraint, we can just add the table to the array and it would work?

yasithA (approved):

services/media/service/src/domains/common/handlers/bulk-edit-item-change-handler.ts
EmperorRXF: yes exactly. there's a clean way to extend the logic for any more tables that meet this criteria

but it anyways requires a change on the `handleDeferrableConstraintTables` - so will leave as-is for now

PR #510, opened 2026-02-24
[AB#53331] style: :arrow_up: upgrade eslint and run prettier independent of eslint

BenjaminSchwendner:
# Pull Request

## Prerequisites

- [x] The PR is targeting the `dev` branch

## Description

- migrating to eslint 9 
- running prettier independent of eslint, following the official recommendations


libs/media-messages/tsconfig.json
```diff
@@ -15,6 +15,7 @@
     "emitDecoratorMetadata": true,
     "experimentalDecorators": true,
     "sourceMap": true,
+    "skipLibCheck": true,
```
EmperorRXF, line 18: I generally try to avoid this - specially on libs we publish. but maybe you had good reason you needed to do this?

EmperorRXF (approved):

libs/media-messages/tsconfig.json
BenjaminSchwendner: If I remember correctly, it was causing issues before, since there were conflicting types from `@types/eslint-scope` and `eslint` package (the former being required by webpack). 
Decided to be a bit more generous on those types.
We can revert it if it causes further issues down the line.

